PR #36, opened 2023-05-18
Fix some shell commands in the K3s API HA setup

atanasdinov:


docs/quickstart/k3s-on-slemicro.md
```diff
@@ -74,6 +74,8 @@ The easiest way to run a K3s HA cluster is by installing a first node using the
 curl -sfL https://get.k3s.io | INSTALL_K3S_EXEC="server --cluster-init --write-kubeconfig-mode=644" K3S_TOKEN=foobar sh -
 ```
 
+**NOTE**: SSH can be utilized in order to easily copy the kube configuration from the first node to the following ones.
```
e-minguez, line 77: I don't get why you are talking about using ssh here, sorry. I mean, you don't need to copy anything from one node to the others, you just need to download the k3s script and point to the API url (with the --server flag) and use the proper token (via the K3S_TOKEN env var).
Would you mind to elaborate? Thanks!

docs/quickstart/k3s-on-slemicro.md
```diff
@@ -130,7 +132,7 @@ The official [kube-vip](https://kube-vip.io/docs/usage/k3s/) documentation expla
 
 ```
 export VIP=192.168.205.10
-cat <<- EOF > /var/lib/rancher/k3s/server/manifests
+cat <<- EOF > /var/lib/rancher/k3s/server/manifests/kube-vip.yaml
```
e-minguez, line 135: Good catch!

docs/quickstart/k3s-on-slemicro.md
```diff
@@ -260,7 +262,7 @@ After a while, the nodes will join the cluster successfully and an HA cluster wi
 The kubeconfig file that is generated as part of the installation has localhost as the Kubernetes API endpoint, so in order to use it from outside, it needs to be changed to the VIP as:
 
 ```
-scp 192.168.205.10:/etc/rancher/k3s/k3s.yaml ~/.kube/config && sed -i '' 's/127.0.0.1/192.168.205.10/g' ~/.kube/config && chmod 600 ~/.kube/config
+scp 192.168.205.10:/etc/rancher/k3s/k3s.yaml ~/.kube/config && sed -i 's/127.0.0.1/192.168.205.10/g' ~/.kube/config && chmod 600 ~/.kube/config
```
e-minguez, line 265: This is actually not wrong :D OSX sed is not GNU sed and it has some differences... so you need to use that trick... I know... Maybe you can add a comment about using GNU sed or put the two versions, one for OSX and the other one for Linux. WDYT?

e-minguez (changes requested):

docs/quickstart/k3s-on-slemicro.md
atanasdinov: Ah, yes, you're completely right. I'll make sure to update it, thanks.

docs/quickstart/k3s-on-slemicro.md
atanasdinov: It seems I've misplaced this. What I meant was that even though connecting new nodes will work by only specifying the address and the token, the agent nodes are missing the admin kube configuration and client requests (e.g. kubectl) will not work for them.

I noticed this while testing the whole setup and used SSH in order to copy the config from the first node to the agent nodes. The agents were then able to do successful operations against the k3s cluster.

Do you think this should be part of the guide or is it irrelevant at this point?

e-minguez: The best practice is to never ssh into the hosts to perform `kubectl` commands (unless something really bad happens) and perform all the operations from the user workstation (I mean, that's why it is an API, not a localhost thing :D) by extracting the kubeconfig file and copying it wherever.

docs/quickstart/k3s-on-slemicro.md
```diff
@@ -263,6 +263,8 @@ The kubeconfig file that is generated as part of the installation has localhost
 scp 192.168.205.10:/etc/rancher/k3s/k3s.yaml ~/.kube/config && sed -i '' 's/127.0.0.1/192.168.205.10/g' ~/.kube/config && chmod 600 ~/.kube/config
 ```
 
+**NOTE:** The above string manipulation will not work for GNU sed and must be replaced with `sed -i 's/127.0.0.1/192.168.205.10/g' ~/.kube/config`.
```
e-minguez, line 266: I'd say it would be better if you can do something similar to https://suse-edge.github.io/dev_howto/create-package-obs#prerequisites (see the source https://raw.githubusercontent.com/suse-edge/suse-edge.github.io/main/docs/dev_howto/create-package-obs.mdx) to have a different tab for each OS instead :)

docs/quickstart/k3s-on-slemicro.md
```diff
@@ -130,7 +130,7 @@ The official [kube-vip](https://kube-vip.io/docs/usage/k3s/) documentation expla
```
e-minguez, line 130: It would be nice if you can add the ipvs instructions as well. Something like:


`lb_enable` [requires](https://kube-vip.io/docs/about/architecture/?query=lb_enable#control-plane-load-balancing) loading the ip_vs modules as:

```
cat <<- EOF > /etc/modules-load.d/ipvs.conf
ip_vs
ip_vs_rr
ip_vs_wrr
ip_vs_sh
nf_conntrack
EOF
```
To avoid rebooting, the modules can be loaded as:
```
for i in $(cat /etc/modules-load.d/ipvs.conf); do modprobe ${i}; done
```

e-minguez (changes requested):

docs/quickstart/k3s-on-slemicro.md
```diff
@@ -124,13 +124,36 @@ The rest of the nodes will be installed after kube-vip as the server URL for the
 
 ### Kube-vip installation
 
-The official [kube-vip](https://kube-vip.io/docs/usage/k3s/) documentation explains the steps in more detail, but essentially it means creating the required resource files for kube-vip to run (RBAC and a DaemonSet) and leveraging [K3s auto-deploy](https://docs.k3s.io/installation/packaged-components#auto-deploying-manifests-addons) feature (aka. any manifest stored in a particular folder of the host `/var/lib/rancher/k3s/server/manifests` will be automatically deployed at the K3s service startup or when the file changes via something similar to `kubectl apply -f`) kube-vip will be executed.
+The official [kube-vip](https://kube-vip.io/docs/usage/k3s/) documentation explains the steps in more detail, but essentially it means creating the required resource files for kube-vip to run (RBAC and a DaemonSet).
+
+**IMPORTANT:** IPVS modules must be loaded in order for the [load balancer feature](https://kube-vip.io/docs/about/architecture/#control-plane-load-balancing) to work.
+This is achieved by creating the following file:
+
+```bash
```
e-minguez, line 132: I guess you need to remove the spaces/tabs at the begining of the lines here

docs/quickstart/k3s-on-slemicro.md
```diff
@@ -124,13 +124,36 @@ The rest of the nodes will be installed after kube-vip as the server URL for the
...
+  ip_vs_sh
+  nf_conntrack
+  EOF
+```
+
+Configurations stored under `/etc/modules-load.d` will be automatically picked up and loaded on boot.
+Loading them for the first time, however, can be achieved without rebooting by executing:
+
+```bash
+  for i in $(cat /etc/modules-load.d/ipvs.conf); do modprobe ${i}; done
```
e-minguez, line 146:
```suggestion
for i in $(cat /etc/modules-load.d/ipvs.conf); do modprobe ${i}; done
```

docs/quickstart/k3s-on-slemicro.md
e-minguez: I'm not 100% positive this will work in plain .md files or they need to be renamed as .mdx... but we can merge this and if it doesn't work, we can rename the file :)

e-minguez (changes requested):

e-minguez (approved):

PR #40, opened 2023-05-24
Update index.md

outrageousbrownies:
Test edit with "Last Updated" date on main page.


e-minguez:
Instead of doing it manually, wouldn't be better to enable the `showLastUpdateTime` setting?  https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs#showLastUpdateTime

PR #45, opened 2023-05-31
Adding ATIP documentation structure

garymackenzie:
Adding structure for ATIP documentation and outline content


sidebars.js
```diff
@@ -51,6 +51,18 @@ const sidebars = {
         }
       ]
     },
+    {
+      "type": "category",
+      "collapsible": true,
+      "collapsed": false,
+      "label": "Documentation",
```
rdoxenham, line 58: Can we rename this to "Product Documentation"?

sidebars.js
```diff
@@ -51,6 +51,18 @@ const sidebars = {
...
+      "items": [
+        {
+          type: 'autogenerated',
```
rdoxenham, line 61: I think that "autogenerated" here means that we end up with "atip" in the menu bar. I'm wondering if it's possible to hard code this to "ATIP" or "Adaptive Telco Infrastructure Platform" so it looks a bit more professional?

docs/product/atip/requirements.md
```diff
@@ -0,0 +1,19 @@
+---
+sidebar_position: 3
+title: Pre-Requisites and Assumptions
```
rdoxenham, line 3: I think this is "Prerequisites"?

docs/product/atip/introduction.md
```diff
@@ -0,0 +1,56 @@
+---
+sidebar_position: 1
+title: Introduction to ATIP Documentation
```
rdoxenham, line 3: Think we can just leave this as "Introduction"

rdoxenham (approved):
LGTM, thanks Gary!

sidebars.js
```diff
@@ -51,6 +51,26 @@ const sidebars = {
...
+    {
+      "type": "category",
+      "collapsible": true,
+      "collapsed": true,
+      "label": "Product Documentation",
+      "items": [
+        {
+          "type": "category",
+          "collapsible": true,
+          "collapsed": true,
```
e-minguez, line 63: Just a nit but I think it would be better if this is false as it requires two clicks to see the actual content.

e-minguez (changes requested):

e-minguez:
LGTM but as it is incomplete, can we mark it as "DRAFT" or something like that?

sidebars.js
garymackenzie: I've made Product Documentation expanded and ATIP collapsed as discussed...

garymackenzie:
> LGTM but as it is incomplete, can we mark it as "DRAFT" or something like that?

Yes, good call - category and all pages now marked as *DRAFT and note added to introduction

e-minguez (approved):

PR #48, opened 2023-06-06
Added homepage, doc cleanup, blog start

jimmyrmartin:
Added the requested homepage, did some cleanup on the doc sidebar so that aren't fully expanded on load. (Cleaner look). Started the work on setting up blogging.


garymackenzie (commented):
I think we'd settled on leaving the top level sections expanded by default to make surfacing content easier at the expense of a cleaner look. @e-minguez had good reasons here.

jimmyrmartin (commented):
Updated the sidebar.js file to be uncollapsed.

e-minguez:
It looks great!

A couple of comments:
* I don't think we need to explicitly talk about Docusaurus in the front page 😅 There is already a mention in the footer.
* I would say to try to be less cheerful. Don't get me wrong on this, I like the mood, but I'll remove some sentences such as the "Hello from ..." or the "We suggest you to please put on your hard hat," 😅  
* Not sure if we want to enable blog or just keep this as documentation.
* Also not sure if we want a front page at all or improve the main doc index page. With this PR it will take an extra click just to go to the main docs, and then another click to read whatever you want to read.

Adding @rdoxenham to the thread :)

rdoxenham:
Thanks for this @jimmyrmartin - so, front page, absolutely this is what we should have. The purpose here should be a landing page where it takes us to the documentation, the blog, video's, and various other sections. Hence why I didn't want the original landing page to simply be the docs page.

e-minguez:
> > 
> 
> I added these per @rdoxenham's asks in the OG Trello/Jira.
> 
> @e-minguez I'm always open to suggestions and viewpoints. The landing page Docusaurus mention is a place holder to make the page not look so dearth of content. It will definitely be replaced with links to the blog, videos, presentations, and so on.
> 

👍🏻 

> I'm trying to add a little personality in our language outside of the technical documentation itself. Documentation that is useful but also friendly (like our Geeko) and playful in small amounts lead to people being more engaged. We may be 'engineers' but we are still people. There is a lot of research in cognitive psychology and documentation on the effect. The documentation part of it, of course can be all buttoned up and fully professional.
> 
> https://www.oreilly.com/content/the-eight-rules-of-good-documentation/
> 
> Again, I'm open to varying opinions, so hopefully this small novel of a "code commit comment" is helpful to explain my thinking.

It is, thanks. From my point of view (this is just me, we can disagree of course!) I prefer to keep the landing page and technical documentation just technical. Bear in mind that there are a lot of non US readers and some of the expressions outside of the technical field can be unknown (I personally need to look for "stalwart" 😄 ). On the other hand we can vary the tone in blog posts of course!

> 
> Also, I won't lie, I still don't fully agree with the uncollapsed sidebar. For someone that deals with a lot of neurodiversity from many angles in my life, it's a bit more to take in and "distracting". Though, I do also see your view on it.
> 

Honestly it is fine if you want to collapse all sections, but based on the fact that we don't have too much content, it serves as a ToC... (see for example https://rancher.github.io/elemental-toolkit/docs/) however when we have enough content I'll encourage to collapse them.

> I'm happy to be working on this with you.

Likewise! 🤗

jimmyrmartin:
TBH: 'Stalwart' might have been a 'ten dollar word' (See side note) pulled out of my mental thesaurus. It seemed to fit at the time of writing. We can iterate, of course.

I do understand the uncollapsed sidebar. I'm not insisting on my way. I see your reasoning over TOC. Just mentioned it due to some personal and professional experience many don't see that side of. For instance, my daughter basically needs a "reading mode" type of experience to be able to focus on the content when being confronted with digital media due to the amount navigational elements vs informational.

Shall we have an Edge Engineering chat about the general content management?

Side note:
'ten dollar word' is likely another American English expression that maybe Rhys or Gary doesn't get, though they are native "English" speakers. I believe it's primarily used these days in the Southern United States. 

https://www.grammarphobia.com/blog/2012/10/ten-dollar-words.html#:~:text=The%20phrase%20“ten%2Ddollar%20word%2C”%20for%20example%2C,to%20important%20or%20pretentious%20words

e-minguez (approved):

e-minguez:
Sorry, I forgot about this yesterday 😓 Merged!

PR #72, opened 2023-09-18
single-iso doc

alknopfler:


docs/dev_howto/create-a-single-iso-image-customized.md
```diff
@@ -0,0 +1,138 @@
+---
+sidebar_position: 3
+title: Create a custom single-iso image (using SLE Micro installer and combustion image) to use it on Virtual CD-ROM
+---
+
+## Clone the SLE Micro installer repository from OBS
+
```
hardys, line 7: Is it worth calling out that this whole OBS section is temporary and won't be required when a future SLEMicro is released with the relevant fixes to combustion?

jdob: +1, this part confused me for a while since I wasn't sure why we were doing it. It took a bit to realize it was just a patch to combustion and not actually anything relevant to my ISO customizations.

docs/dev_howto/create-a-single-iso-image-customized.md
```diff
@@ -0,0 +1,138 @@
...
+xorriso -indev ./SLE-Micro.x86_64-5.4.0-Default-SelfInstall-Build15.1.install.iso \
+        -outdev ./SLE-Micro-Selfinstall-with-mycombustion-single-iso.iso \
+        -map ~/my-local-path/combustion /combustion \
+        -boot_image any replay -changes_pending yes
+```
+
+After that, we should have the final iso image with the combustion script included `SLE-Micro-Selfinstall-with-mycombustion-single-iso.iso`
+
+
+> This feature will be added to the SLE Micro 5.5 release, but meanwhile, we could use this workaround to create a single-iso image with the combustion script included.
```
jdob, line 138: This might make more sense to be up higher so the reader isn't unsure of why the combustion packaging steps are needed.

jdob (changes requested):
Only major request is to confirm the OBS link. The others are suggestions.

docs/dev_howto/create-a-single-iso-image-customized.md
alknopfler: I added at the end of the howto, but I agree that it's better to explain why we need to do this until we release 5.5 with the modifications included.

docs/dev_howto/create-a-single-iso-image-customized.md
```diff
@@ -0,0 +1,138 @@
+---
+sidebar_position: 3
+title: Create a custom single-iso image (using SLE Micro installer and combustion image) to use it on Virtual CD-ROM
+---
+
+## Clone the SLE Micro installer repository from OBS
+
+* Log in to [OBS](https://build.opensuse.org)
+* Go to the [SLE Micro installer repository](https://build.opensuse.org/project/show/SUSE:SLE-15-SP4:Update:Products:Micro54/SLE-Micro)
```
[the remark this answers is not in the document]
alknopfler: you're right, the last /SLE-Micro not needed. I will change it

docs/dev_howto/create-a-single-iso-image-customized.md
alknopfler: yes, that's my first comment with Steve. I think I need to explain before why we need it.
Thanks

docs/dev_howto/create-a-single-iso-image-customized.md
```diff
@@ -0,0 +1,145 @@
+---
+sidebar_position: 3
+title: Create a custom single-iso image (using SLE Micro installer and combustion image) to use it on Virtual CD-ROM
+---
+
+The feature to create a single iso (installer image and combustion) to be used by Virtual CD-Rom, will be added to the SLE Micro 5.5 release.
+When the SLE Micro 5.5 is released, we will be able to create a single-iso image with the combustion script included just following the last part of this document [here](#download-the-new-iso-image-to-prepare-it-with-xorriso-and-adding-combustion).
```
jdob, line 7:
```suggestion
When the SLE Micro 5.5 is released, or if you want to use the existing [build](https://build.opensuse.org/project/show/home:amorgante:branches:SUSE:SLE-15-SP4:Update:Products:Micro54) created in this guide, we will be able to create a single-iso image with the combustion script included just following the last part of this document [here](#download-the-new-iso-image-to-prepare-it-with-xorriso-and-adding-combustion).
```

jdob (approved):
LGTM to land with or without my suggestion.

PR #73, opened 2023-09-19
Adding instructions for NVIDIA GPU's on SLE Micro

rdoxenham:


docs/quickstart/nvidia-slemicro.md
```diff
@@ -0,0 +1,318 @@
...
+
+It's important to call out that the support for these drivers is provided by both SUSE and NVIDIA in tight collaboration, however if you have any concerns or questions about the combination in which you're utilising the drivers, then please ask your SUSE or NVIDIA account managers for further assistance. If you're planning on utilising [NVIDIA AI Enterprise](https://www.nvidia.com/en-gb/data-center/products/ai-enterprise/) (NVAIE) you will need to ensure that you're using an [NVAIE certified GPU](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/platform-support.html#supported-nvidia-gpus-and-systems), which *may* require the use of proprietary NVIDIA drivers. If you're unsure, please speak with your NVIDIA representative.
+
+> NOTE: The instructions provided below will drastically simplify in the coming weeks; the instructions below demonstrate how to pull some packages from SLES repos and install them on SLE Micro, and we're currently porting these packages over to SLE Micro 5.3, 5.4, and 5.5. They'll also be available in the ALP-based Micro equivalent in the coming months.
```
alknopfler, line 14: I'd move to the end but it's a personal opinion

docs/quickstart/nvidia-slemicro.md
```diff
@@ -0,0 +1,318 @@
...
+```shell
+exit
+```
+
+> NOTE: Please make sure that you've exited the `transactional-update` session before proceeding!
+
+Next, if you're *not* using a supported GPU, remembering that the list can be found [here](https://github.com/NVIDIA/open-gpu-kernel-modules#compatible-gpus), you can see if the driver will work by enabling support at the module level, but your mileage may vary -- skip this step if you're using a *supported* GPU:
+
+```shell
+sed '/NVreg_OpenRmEnableUnsupportedGpus/s/^#//g' /etc/modprobe.d/50-nvidia-default.conf
```
e-minguez, line 105:
```suggestion
sed -i '/NVreg_OpenRmEnableUnsupportedGpus/s/^#//g' /etc/modprobe.d/50-nvidia-default.conf
```

`-i` is required for inline changes, otherwise it will output the content to stdout.

docs/quickstart/nvidia-slemicro.md
```diff
@@ -0,0 +1,318 @@
...
+
+> NOTE: The instructions provided below will drastically simplify in the coming weeks; the instructions below demonstrate how to pull some packages from SLES repos and install them on SLE Micro, and we're currently porting these packages over to SLE Micro 5.3, 5.4, and 5.5. They'll also be available in the ALP-based Micro equivalent in the coming months.
+
+## Prerequisites
+
+If you're following this guide, it's assumed that you've got the following already available:
+
+* At least one host with SLE Micro 5.3+ installed; this can be physical or virtual.
+* Your host(s) is/are attached to a subscription as this will be required for package access - an evaluation is available [here](https://www.suse.com/download/sle-micro/).
```
alknopfler, line 21: maybe better registered on SCC???
```suggestion
* Your host(s) is/are registered on the Suse Customer Center `SCC` as this will be required for package access - an evaluation is available [here](https://www.suse.com/download/sle-micro/).
```

alknopfler (commented):
small changes!!! LGTM

docs/quickstart/nvidia-slemicro.md
```diff
@@ -0,0 +1,318 @@
...
+
+Once this has been installed successfully, don't exit from the container, we'll run the `deviceQuery` CUDA example, which will comprehensively validate GPU access via CUDA, and from within the container itself:
+
+```shell
+/usr/local/cuda-12/extras/demo_suite/deviceQuery
+```
+
+If successful, you should see output that shows similar to the following, noting the `Result = PASS` message at the end of the command:
+
+```shell
```
diconico07, line 225: nit: this is not really a shell snippet, raw text should be a better choice here

docs/quickstart/nvidia-slemicro.md
```diff
@@ -0,0 +1,318 @@
...
+|=======================================================================================|
+|  No running processes found                                                           |
++---------------------------------------------------------------------------------------+
+```
+
+...and that's it! You've successfully installed and verified that the NVIDIA drivers are loaded into SLE Micro.
+
+## Further Validation
+
+At this stage, all we've been able to verify is that at the host level the NVIDIA device can be accessed and that the drivers are loading successfully. However, if we want to be sure that it's functioning, a simple test would be to try and validate that the GPU can take instruction from a user-space application, ideally via a container, and through the CUDA library, as that's typically what a real workload would utilise. For this, we can make a further modification to the host OS by installing the `nvidia-container-toolkit`. First, open up another `transactional-update` shell, noting that we could have done this in a single transaction in the previous step, but to many (e.g. customers wanting to use Kubernetes) this step won't be required:
```
diconico07, line 153: Maybe talk or link about how to validate with kubernetes here as you mention this is only for those who don't use kubernetes

docs/quickstart/nvidia-slemicro.md
```diff
@@ -0,0 +1,318 @@
...
+  Compute Mode:
+     < Default (multiple host threads can use ::cudaSetDevice() with device simultaneously) >
+> Peer access from NVIDIA A100-PCIE-40GB (GPU0) -> NVIDIA A100-PCIE-40GB (GPU1) : Yes
+> Peer access from NVIDIA A100-PCIE-40GB (GPU1) -> NVIDIA A100-PCIE-40GB (GPU0) : Yes
+
+deviceQuery, CUDA Driver = CUDART, CUDA Driver Version = 12.2, CUDA Runtime Version = 12.1, NumDevs = 2, Device0 = NVIDIA A100-PCIE-40GB, Device1 = NVIDIA A100-PCIE-40GB
+Result = PASS
+```
+
+From here, you can continue to run any other CUDA workload - you can utilise compilers, and any other aspect of the CUDA ecosystem to run some further tests. When you're done you can exit from the container, noting that whatever you've installed in there is ephemeral (so will be lost!), and hasn't impacted the underlying operating system:
```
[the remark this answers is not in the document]
rdoxenham: Thanks Alberto, but "noting" is the correct word here, like "take a note"

docs/quickstart/nvidia-slemicro.md
rdoxenham: Yeah, I've got a section at the bottom that I want to fill out, but I don't want to delay this first release :)

PR #78, opened 2023-09-25
Document Cluster API core concepts

atanasdinov:


e-minguez (approved):
It would be nice to add some examples but I'll merge it anyway and maybe for a follow up PR :)

PR #86, opened 2023-11-07
Removed the requirement for toolbox for SLE Micro

rdoxenham:
NVIDIA drivers.


docs/quickstart/nvidia-slemicro.md
```diff
@@ -82,29 +41,28 @@ zypper ar https://developer.download.nvidia.com/compute/cuda/repos/sles15/x86_64
 zypper ar https://download.nvidia.com/suse/sle15sp4/ nvidia-sle15sp4-main
 ```
 
-Next, move to the `/root/nvidia` directory, which will contain a couple of directories that the previous `zypper` command created with the downloaded packages from SLES. Then, we can request that the package manager install the pre-built signed kernel modules, the firmware packages, and the additional useful utilities package:
+You can then install the driver and the `nvidia-compute-utils` for additional utilities:
 
 ```shell
-cd /root/nvidia/container-suseconnect-zypp:SLE-Module-Basesystem15-SP4-Updates/x86_64
-zypper in nvidia-open-driver-G06-signed-kmp-default* kernel-firmware-nvidia-gspx-G06* nvidia-compute-utils-G06
+zypper install -y nvidia-open-driver-G06-signed-kmp=535.86.05 kernel-firmware-nvidia-gspx-G06 nvidia-compute-utils-G06
```
e-minguez, line 47: is the version really needed or will it work without it?

rdoxenham: A version is added specifically here because there's unfortunately a mismatch between the NVIDIA repo and the SLE Micro repo. We're usually a little more aggressive with pushing updates, and it's possible that customers will install a newer package and the matching proprietary software from the NVIDIA side isn't yet available. Therefore the documentation specifically says to check this and match the version

Kristian-ZH (approved):

PR #97, opened 2024-01-03
Rough draft of Elemental and EIB quickstarts

agracey:


docs/quickstart/elemental.mdx
```diff
@@ -0,0 +1,400 @@
...
+- (If using physical hardware) A USB flash drive to boot from
+  - Existing will be overwritten as part of the process
+
+In the writing of this guide, I will be using a Digital Ocean droplet to host the upstream cluster and an Intel NUC as the downstream device. For building the installation media, I will be using my desktop running [openSUSE Tumbleweed](https://get.opensuse.org/tumbleweed/).
+
+## How to Use Elemental
+
+The basic steps to install and use Elemental are:
+
+- Build bootstrap cluster
```
hardys, line 48: We're mixing terminology here with upstream cluster, bootstrap cluster etc - can we standardise perhaps on "Management Cluster" ?

Bootstrap has a special meaning in the CAPI flow, but I think "management" should be Ok to use for all 3 flows without causing confusion?

docs/quickstart/elemental.mdx
```diff
@@ -0,0 +1,400 @@
...
+
+## What is Elemental?
+
+Elemental is a software stack that provides for remote control over the lifecycles of the Operating System, Kubernetes, and workload in a secure and declarative way. 
+
+
+
+## Why use this method
+
+Elemental can be very useful for scenarios where the devices that you want to control are not on the same network as the upstream cluster or don't have a BMC onboard to allow more direct control.
```
hardys, line 16: I would personally not start with the same-network constraint, while that's true it's not the defining characteristic of this flow, it's that it provides a "boot it yourself" flow where an image can be booted, then hosts can be onboarded and bound to create clusters.

How about something like "Elemental provides an onboarding flow where hosts can be booted via a discovery image, onboarded into an inventory, then bound to create clusters.  This is very useful where a "boot it yourself" flow is required, such as in cases where the target systems do not support out-of-band management, or are on a remote network not directly accessible by the management cluster"

docs/quickstart/elemental.mdx
agracey: While I personally agree with "management", I think we might want to go with "upstream" since that's what the Rancher docs use. Thoughts?

agracey: I don't know why I picked "bootstrap" for this specific instance...

docs/quickstart/elemental.mdx
```diff
@@ -0,0 +1,400 @@
...
+#### Setup DNS
+
+Before continuing, you will need to setup access to your cluster. As with the setup of the cluster itself, how you configure DNS will be different depending on where it's being hosted.
+
+:::tip 
+If you don't want to deal with setting up DNS records (for example this is just an ephemeral test server), you can use a service like [sslip.io](https://sslip.io) instead. With this service, you can resolve any ip address with `<address>.sslip.io`. 
+:::
+
+#### Install Rancher
+
```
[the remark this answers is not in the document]
agracey: Yeah, I need to figure out importing content in docusaurus.

docs/quickstart/elemental.mdx
hardys: Sounds good to me - aligning with the Rancher docs wherever possible absolutely makes sense - and perhaps we can link some relevant sections like https://ranchermanager.docs.rancher.com/reference-guides/rancher-manager-architecture/architecture-recommendations#separation-of-rancher-and-user-clusters for example (a diagram will also help - we should chat about how to make those consistent)

agracey:
Merging without waiting for review to allow for easier collaboration.

PR #117, opened 2024-03-13
Add Longhorn documentation

dprodanov4:


asciidoc/components/longhorn.adoc
```diff
@@ -10,4 +10,195 @@ ifdef::env-github[]
 :warning-caption: :warning:
 endif::[]
 
-https://longhorn.io/docs/
+Longhorn is a lightweight, reliable, and user-friendly distributed block storage system designed for Kubernetes.
+As an open-source project, Longhorn was initially developed by Rancher Labs and is currently incubated under the CNCF.
+
+== Prerequisites
+
+* K3s installed. You can find a guide on how to install it on SLE Micro https://documentation.suse.com/trd/kubernetes/single-html/kubernetes_ri_k3s-slemicro/index.html#id-k3s[here]
```
agracey, line 18: Or RKE2

asciidoc/components/longhorn.adoc
```diff
@@ -10,4 +10,195 @@ ifdef::env-github[]
...
+* Open-iSCSI
+** You can install it on SLE Micro by running `sudo transactional-update pkg install open-iscsi` and rebooting the host.
```
agracey, line 20: Is this still needed? For some reason I thought they had simplified this?

Also, should we suggest using EIB to install open-iscsi?

agracey (approved):
Thank you for this! 

I added a few comments/questions but I'm fine with merging as is if you would prefer to address those during editing.

asciidoc/components/longhorn.adoc
dprodanov4: Unfortunately, it is still needed I tried on SLE Micro and Ubuntu in both cases it was required. In the longhorn docs 1.6.0 it is still specified as a requirement on each node to have it installed.

asciidoc/components/longhorn.adoc
dprodanov4: Will update it now.

asciidoc/components/longhorn.adoc
dprodanov4: Regarding the EIB, will be testing with it today.

dprodanov4:
Will merge this for now and add the change for RKE2 in the edits together with new section about how to use EIB for the deployment.

PR #131, opened 2024-03-25
Update rancher.adoc

atrendafilov:


agracey (approved):
Thank you for this! 

The content is good but I'm a bit concerned about readability once rendered. I'm going to go ahead and merge so I can more easily render locally then clean up if needed

PR #132, opened 2024-03-25
Updating repo readme to reflect change to asciidoc

agracey:

PR #135, opened 2024-03-25
Reinstate deploy workflows

hardys:
These were removed in #109 so revert that here and adjust to build the asciidoc format

Based on the steps described in #132

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,47 @@
+name: Deploy to GitHub Pages
+
+# https://github.com/peaceiris/actions-gh-pages/issues/852
+permissions:
+  contents: write
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+    # Review gh actions docs if you want to further define triggers, paths, etc
+    # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#on
+
+jobs:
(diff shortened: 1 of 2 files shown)


hardys:
This was tested via my fork and we can see https://github.com/hardys/suse-edge.github.io/tree/gh-pages worked, I've not configured the actual site to preview yet though

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,47 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Run the docs build in a container
+        uses: addnab/docker-run-action@v3
+        with:
+            image: registry.opensuse.org/home/atgracey/cnbp/containers/builder:latest
```
hardys, line 27: @agracey this image seems very large, is there anything we can do to improve that?  Probably we should tag a version other than latest too and perhaps move somewhere other than a home project?

agracey: Waiting on a build but I believe this will be smaller once it builds: https://build.opensuse.org/package/show/home:atgracey:utilities/daps

It doesn't have the extra buildpack bits. 

I think I remember Frank saying they might have this image already so I'll check with him and let you know.

agracey (approved):

PR #146, opened 2024-04-04
Update to elemental quick start

dbw7:
I updated the elemental quickstart to fix a few errors during deployment and also to make the process more point and click. The only thing I want to add is a note about the TPM hardware needing to be added but I'm not sure where to put that, or if I even should. Eventually would also like to clean it up and add more detail on the deployment + testing but will save that for later.


dbw7:
Also, I'm not sure if I messed up the formatting so please let me know

agracey (approved):
Thank you so much for reviewing and fixing this!

PR #147, opened 2024-04-05
Add documentation for the Endpoint Copier Operator

Kristian-ZH:
Add documentation for the Endpoint Copier Operator

### DIFF
--- a/asciidoc/components/endpoint-copier-operator.adoc
+++ b/asciidoc/components/endpoint-copier-operator.adoc
@@ -1,0 +1,29 @@
+[#components-eco]
+= Endpoint Copier Operator
+
+ifdef::env-github[]
+:imagesdir: ../images/
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:important-caption: :heavy_exclamation_mark:
+:caution-caption: :fire:
+:warning-caption: :warning:
+endif::[]
+
+https://github.com/suse-edge/endpoint-copier-operator[Endpoint Copier Operator] is a Kubernetes operator whose purpose is to create a copy of a Kubernetes Service and Endpoint and to keep them synced.
+
+== How does SUSE Edge use Endpoint Copier Operator?
+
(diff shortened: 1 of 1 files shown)


asciidoc/components/endpoint-copier-operator.adoc
```diff
@@ -0,0 +1,29 @@
...
+:warning-caption: :warning:
+endif::[]
+
+https://github.com/suse-edge/endpoint-copier-operator[Endpoint Copier Operator] is a Kubernetes operator whose purpose is to create a copy of a Kubernetes Service and Endpoint and to keep them synced.
+
+== How does SUSE Edge use Endpoint Copier Operator?
+
+At SUSE Edge, the Endpoint Copier Operator plays a crucial role in achieving High Availability (HA) setup for K3s/RKE2 clusters. This is accomplished by creating a `kubernetes-vip` service of type `LoadBalancer`, ensuring its Endpoint remains in constant synchronization with the kubernetes Endpoint. <<components-metallb,MetalLB>> is leveraged to manage the `kubernetes-vip` service, as the exposed IP address is used from other nodes to join the cluster.
+
+== Best Practices
```
agracey, line 19: This empty section will cause the docs to not build. Let's go ahead and pull it out for now until we have best practices defined

agracey (changes requested):
Thank you for writing this! I can't believe I forgot to add it... 

Only one change needed so it can render correctly

asciidoc/components/endpoint-copier-operator.adoc
agracey: Well maybe not? It seem like the validation passed... I'm confused because I've had to add place holder text in a few other places.

Kristian-ZH: It is not actually Empty, we have `Getting started` Inner section in it.

Also other components like k3s, EIB ... use the same structure

agracey: Ah... I'm apparently blind this morning because the 3 "="s looked like 2

agracey (approved):

PR #148, opened 2024-04-05
update metallb manifests

alknopfler:
- update the metallb manifests to reduce complexity and update the chart links


atanasdinov (approved):

agracey (approved):
This looks a lot simpler! 

We should link to the endpoint operator page that's being added with https://github.com/suse-edge/suse-edge.github.io/pull/147

PR #157, opened 2024-04-12
ATIP management cluster fix hostname script issue

hardys:
/etc/hostname is only populated when the hostname is statically configured, if the hostname is provided via DHCP this won't work so we can use `hostname` instead


hardys:
This was discovered by @peteingithub  during testing (thanks!)

@alknopfler @e-minguez - I guess we'll want to rework this section soon to align with the EIB helm support, but should we land this in the meantime?

e-minguez (approved):

alknopfler (approved):

PR #167, opened 2024-04-22
Adding version matrix partial

agracey:


asciidoc/edge-book/version-matrix.adoc
```diff
@@ -0,0 +1,23 @@
+[#component-version-matrix]
+.Supported Component Versions
+[options="header"]
+|======
+| Name | Version 
+| SLE Micro | 5.5
+| Rancher Prime | 2.8.3
+| Fleet | 
+| K3s | 
+| RKE2 |
```
atanasdinov, line 10: Both of these should be 1.28.8 which is the currently supported version of [Rancher 2.8.3](https://github.com/rancher/rancher/releases/tag/v2.8.3).

asciidoc/edge-book/version-matrix.adoc
```diff
@@ -0,0 +1,23 @@
...
+| RKE2 | 
+| Metal^3^ | 0.6.5
+| MetalLB | 0.14.3
+| Elemental | 1.4.0
+| Edge Image Builder | 1.0.1
+| NM Configurator | 0.2.3
+| Longhorn | 1.6.1
+| NeuVector| 5.3.0
+| KubeVirt | 0.2.4
+| Containerized Data Importer | 0.2.3
```
atanasdinov, line 19: These are confusing because they're the versions of the charts. The actual KubeVirt and CDI versions are 1.1.1 and 1.58.0 respectively.

agracey: Good catch

asciidoc/edge-book/version-matrix.adoc
agracey: Thank you!

agracey:
Marking as ready for review but I have an open question about how to show a subtree under Appendix. @fsundermeyer is this possible?

hardys:
> Aren't we missing Akri?

I believe Akri is a tech-preview for 3.0, so while we could add it to this list we should add some kind of note to reflect that

atanasdinov (approved):

asciidoc/edge-book/version-matrix.adoc
```diff
@@ -0,0 +1,36 @@
...
+.Supported Component Versions
+[options="header"]
+|======
+| Name | Version 
+| SLE Micro | 5.5
+| Rancher Prime | 2.8.3
+| Fleet | 0.9.2
+| K3s | 1.28.8
+| RKE2 | 1.28.8
+| Metal^3^ | 0.6.5
```
hardys, line 24: We probably need to clarify what the version refers to, because currently we're mixing upstream release versions (e.g kubevirt which is upstream 1.1.1 but chart version 0.2.4) with downstream chart versions (e.g metal3 which is downstream chart version 0.6.5 but contains several upstream released versions)

I'm not sure how best to handle that - perhaps we add another column with "Chart Version" that can be n/a for non-chart things, then a Version/Notes column which includes details of the upstream releases?

agracey: Good call.

agracey:
> Aren't we missing Akri?

Yeah, I left it off because I wasn't sure how to represent it. Maybe I put "latest" as the version?

fsundermeyer:
> Marking as ready for review but I have an open question about how to show a subtree under Appendix. @fsundermeyer is this possible?

What exactly do you mean by "subtree"? 

All .adoc files in a subdirectory? There is no way to _automatically_ include all files in a subdirectory. You either have to specify them one by one in `edge.adoc` or create an include file in the subdirectory comprising all files and thenm include that file in `edge.adoc`.

agracey:
When we have more than one Appendix, it could be nice to see all the appendix in a tree on the navigation sidebar instead of all being top level.

PR #168, opened 2024-04-23
Add Edge Networking component

atanasdinov:


asciidoc/components/networking.adoc
```diff
@@ -0,0 +1,677 @@
...
+:important-caption: :heavy_exclamation_mark:
+:caution-caption: :fire:
+:warning-caption: :warning:
+endif::[]
+
+In this section we’re going to describe the approach SUSE Edge undertakes to customize the various different networking options on your edge nodes.
+Modern Linux distributions (incl. SLE Micro) are leveraging NetworkManager to suit their configuration needs.
+
+At the time of this writing, SUSE Edge Networking is not a standalone product. This guide will, however, dive deep into
+what it consists of and how it is integrated into the SUSE Edge solution.
```
hardys, line 17: This introduction can be simplified, I'd suggest replacing both paragraphs with something like:

```
This section describes the approach to network configuration in the SUSE Edge solution.  We will show how to configure NetworkManager on SLEMicro in a declarative manner, and explain how they are integrated.
```

asciidoc/components/networking.adoc
```diff
@@ -0,0 +1,677 @@
...
+
+== NetworkManager overview
+
+NetworkManager is a program that manages the primary network connection and other connection interfaces.
+It is handled by systemd and is shipped with all necessary service unit files.
```
hardys, line 22: I think we could simplify this to e.g "NetworkManager is a tool that manages the primary network connection and other connection interfaces."

Since it's not a single program but a combination of components, and the system detail is not specific to NM.

asciidoc/components/networking.adoc
rdoxenham: Agree, and anything capitalised gives the impression that it's a product, so rather than "SUSE Edge Networking", I'd make it something like "network customisation options available in SUSE Edge"

asciidoc/components/networking.adoc
```diff
@@ -0,0 +1,677 @@
...
+
+NetworkManager stores all network configurations as connections, which are collections of data that describes how to create or connect to a network.
+These connections are stored as files in the `/etc/NetworkManager/system-connections/` directory.
+
+Details about NetworkManager can be found in the https://documentation.suse.com/sle-micro/5.5/html/SLE-Micro-all/cha-nm-configuration.html[upstream SLE Micro documentation].
+
+== nmstate overview
+
+There are different solutions attempting to automate and manage host networking and one of the most popular ones is nmstate.
+It is a library (with an accompanying CLI tool) which offers a declarative API for network configurations via a predefined schema.
```
hardys, line 32: I would remove the first sentence -  we just should explain what the component does.

Or if we want to clarify that it's a widely adopted API we could say "It is a widely adopted library ..." - that's not quite the same as stating an opinion regarding popularity :)

asciidoc/components/networking.adoc
```diff
@@ -0,0 +1,677 @@
...
+
+At the time of this writing, SUSE Edge Networking is not a standalone product. This guide will, however, dive deep into
+what it consists of and how it is integrated into the SUSE Edge solution.
+
+== NetworkManager overview
+
+NetworkManager is a program that manages the primary network connection and other connection interfaces.
+It is handled by systemd and is shipped with all necessary service unit files.
+
+NetworkManager stores all network configurations as connections, which are collections of data that describes how to create or connect to a network.
```
rdoxenham, line 24: "network configurations as connection files that store the desired state".

asciidoc/components/networking.adoc
```diff
@@ -0,0 +1,677 @@
...
+These connections are stored as files in the `/etc/NetworkManager/system-connections/` directory.
+
+Details about NetworkManager can be found in the https://documentation.suse.com/sle-micro/5.5/html/SLE-Micro-all/cha-nm-configuration.html[upstream SLE Micro documentation].
+
+== nmstate overview
```
rdoxenham, line 29: Maybe "Overview of nmstate"? Seems a bit off not capitalising the first letter here.

asciidoc/components/networking.adoc
```diff
@@ -0,0 +1,677 @@
...
+
+There are different solutions attempting to automate and manage host networking and one of the most popular ones is nmstate.
+It is a library (with an accompanying CLI tool) which offers a declarative API for network configurations via a predefined schema.
+
+Details about nmstate can be found in the https://nmstate.io/[upstream documentation].
+
+== Enter: NetworkManager Configurator (nmc)
+
+Edge Networking is achieved via a CLI tool called NetworkManager Configurator or _nmc_ in short.
```
rdoxenham, line 38: Same here regarding "Networking" being capitalised.

rdoxenham: Also "for short" rather than "in short".

asciidoc/components/networking.adoc
```diff
@@ -0,0 +1,677 @@
...
+This tool allows us to generate network configurations from predefined desired states and to apply those across many different nodes in an automated fashion.
+
+Details about the NetworkManager Configurator (nmc) can be found in the https://github.com/suse-edge/nm-configurator[upstream repository].
+
+== How does SUSE Edge use NetworkManager Configurator?
+
+SUSE Edge utilizes _nmc_ for the network customizations in the various different provisioning models:
+
+* Custom network configurations in the Direct Network Provisioning scenarios (<<quickstart-metal3>>)
+* Declarative static configurations in the Image Based Provisioning scenarios (<<quickstart-eib>>)
```
rdoxenham, line 50: We also would support this flow via the Elemental registration approach, which is akin to image based, but not unique to it.

asciidoc/components/networking.adoc
```diff
@@ -0,0 +1,677 @@
...
+
+== Configuring with Edge Image Builder
+
+Edge Image Builder is a staple when it comes to configuring multiple hosts with a single OS image.
+In this section we'll show how you can use a declarative approach to describe the desired network states,
+how those are converted to the respective NetworkManager connections and are then applied during the provisioning process.
```
rdoxenham, line 56:
```suggestion
how those are converted to the respective NetworkManager connections, and are then applied during the provisioning process.
```

asciidoc/components/networking.adoc
```diff
@@ -0,0 +1,677 @@
...
+
+=== Prerequisites
+
+If you're following this guide, it's assumed that you've got the following already available:
+
+* An x86_64 physical host (or virtual machine) running SLES 15 SP5 or openSUSE Leap 15.5,
+and with virtualization extensions enabled in the BIOS
+(see https://documentation.suse.com/sles/15-SP5/html/SLES-all/cha-virt-support.html#sec-kvm-requires-hardware[here] for details)
```
rdoxenham, line 64: Why do we need virtualisation here?

asciidoc/components/networking.adoc
```diff
@@ -0,0 +1,677 @@
...
+        - ip: 192.168.122.55
+          prefix-length: 24
+      dhcp: false
+      enabled: true
+    ipv6:
+      enabled: false
+EOF
+----
+
+In this example we define a desired state of two Ethernet interfaces (eth0 and eth3),
```
rdoxenham, line 208: Whilst this is technically feasible, it just seems a bit strange to use "eth0" and "eth3" in the documentation. If possible, can we stick with "eth0" and "eth1"?

asciidoc/components/networking.adoc
```diff
@@ -0,0 +1,677 @@
...
+their requested IP addresses, routing and DNS resolution.
```
rdoxenham, line 209:
```suggestion
their requested IP addresses, routing, and DNS resolution.
```

asciidoc/components/networking.adoc
```diff
@@ -0,0 +1,677 @@
...
+
+[WARNING]
+====
+You must ensure that the MAC addresses of all Ethernet interfaces are listed.
+Those are used during the provisioning process to determine which configurations are relevant for the particular node.
```
rdoxenham, line 214: We likely need to make it clear that the MAC address is used to uniquely identify which host is which, and we do this to enable a single ISO to be used for multiple hosts.

asciidoc/components/networking.adoc
```diff
@@ -0,0 +1,677 @@
...
+----
+
+As previously mentioned, the NetworkManager Configurator (_nmc_) tool expects an input in the form of predefined schema.
+You can find how to set up a wide variety of different networking options in the https://nmstate.io/examples.html[upstream NMState examples documentation].
+
+This guide will explain how to configure the networking on three different nodes:
+
+* A node which uses two Ethernet interfaces
+* A node which uses network bonding
+* A node which uses a network bridge
```
rdoxenham, line 154: Make it clear here that this is not a recommended production approach, but merely used as an example to show various different options that can be configured. We would recommend that networking configurations are homegeneous amongst nodes, or at least amongst roles within a given cluster.

asciidoc/components/networking.adoc
```diff
@@ -0,0 +1,677 @@
...
+Kubernetes ................... [SKIPPED]
+Certificates ................. [SKIPPED]
+Building RAW image...
+Kernel Params ................ [SKIPPED]
+Image build complete!
+----
+
+The snippet above tells us that the `Network` component has successfully been configured, and we can proceed with provisioning our edge nodes.
+
+> NOTE: A log file (`network-config.log`) and the respective NetworkManager connection files can be inspected in the resulting `_build` directory.
```
rdoxenham, line 387: ...along with a timestamped directory for the image run.

asciidoc/components/networking.adoc
```diff
@@ -0,0 +1,677 @@
...
+
+Let's copy the resulting RAW image:
+
+[,shell]
+----
+mkdir edge-nodes && cd edge-nodes
+cp $CONFIG_DIR/modified-image.raw node1.raw
+cp $CONFIG_DIR/modified-image.raw node2.raw
+cp $CONFIG_DIR/modified-image.raw node3.raw
+cp $CONFIG_DIR/modified-image.raw node4.raw
```
rdoxenham, line 399: We could likely simply this, e.g.

```
for i in {1..4}; do cp $CONFIG_DIR/modified-image.raw node$i.raw; done
```

asciidoc/components/networking.adoc
```diff
@@ -0,0 +1,677 @@
...
+In these cases, we will default to DHCP to configure the network interfaces.
+
+Let's create the virtual machine:
+
+[,shell]
+----
+virt-install --name node4 --ram 10000 --vcpus 6 --disk path=node4.raw,format=raw --osinfo detect=on,name=sle-unknown --graphics none --console pty,target_type=serial --network default --virt-type kvm --import
+----
+
+Once the virtual machine is up and running, we can confirm that this node is using a random IP address for its network interface:
```
rdoxenham, line 649: We should likely show the output from the Combustion run which shows that `nmc` failed to match the host?

asciidoc/components/networking.adoc
```diff
@@ -0,0 +1,677 @@
...
+== How does SUSE Edge use NetworkManager Configurator?
+
+SUSE Edge utilizes _nmc_ for the network customizations in the various different provisioning models:
+
+* Custom network configurations in the Direct Network Provisioning scenarios (<<quickstart-metal3>>)
+* Declarative static configurations in the Image Based Provisioning scenarios (<<quickstart-eib>>)
+
+== Configuring with Edge Image Builder
+
+Edge Image Builder is a staple when it comes to configuring multiple hosts with a single OS image.
```
hardys, line 54: I would phrase this as "Edge Image Builder (EIB) is a tool which enables configuring multiple hosts with a single OS image."

That makes the phrasing slightly clearer IMO, and introduces the acronym which is used in the subsequent sections.

asciidoc/components/networking.adoc
```diff
@@ -0,0 +1,677 @@
...
+apiVersion: 1.0
+image:
+  arch: x86_64
+  imageType: raw
+  baseImage: SLE-Micro.x86_64-5.5.0-Default-GM.raw
+  outputImageName: modified-image.raw
+operatingSystem:
+  users:
+    - username: root
+      encryptedPassword: $6$jHugJNNd3HElGsUZ$eodjVe4te5ps44SVcWshdfWizrP.xAyd71CVEXazBJ/.v799/WRCBXxfYmunlBO2yp1hm/zb4r8EmnrrNCF.P/
```
hardys, line 122: Should we add an inline comment or an admonition showing how to generate the password?

asciidoc/components/networking.adoc
```diff
@@ -0,0 +1,677 @@
...
+EOF
+----
+
+The `image` section is required, and it specifies the input image, its architecture and type, as well as what the output image will be called.
+The `operatingSystem` section, albeit not necessary, will help us to log in on the provisioned systems with the `root:eib` credentials pair.
```
hardys, line 127: I would phrase this as

The operatingSystem` section is optional, and contains configuration to enable login on the provisioned systems with the `root/eib` username/password.

hardys (commented):
Overall looks great, a few minor comments added

asciidoc/components/networking.adoc
atanasdinov: For the provisioning examples using virt-install below. How should we handle this?

asciidoc/components/networking.adoc
atanasdinov: It's to emphasise on what happens when the actual NIC is eth1 rather than eth3. Any better way of doing it?

hardys: Not exactly related, but we should perhaps add something about the option to use predictable nic names instead of ethX where there's a risk the device/nic mapping is not persistent over time?

Perhaps that's less of an issue in the case where we mandate a MAC to pin the nic to a device, but in the Metal3 flow we describe using `net.ifnames=1` in the EIB kernelArgs to enable predictable names instead of the default ethX (IMO this is a better default, but particularly important for telco use-cases where there are often multiple nics)

https://suse-edge.github.io/atip-automated-provisioning.html#ztp-eib-edge-image

agracey: Is there a reason we don't default to ifnames? Is it just that we are all more comfortable with eth0?

agracey: And if it a comfort thing, I totally get it 😄

atanasdinov: We don't modify any kernel args by default currently. `net.ifnames=1` must be explicitly enabled. We can consider it in future releases, though.

asciidoc/components/networking.adoc
rdoxenham: I think we should make it clear in the docs that this is a practical example that can be followed for explanation purposes, not how you would operate in a production environment. As otherwise we'll have folks assuming that you need virtualisation to run EIB, etc.

asciidoc/components/networking.adoc
```diff
@@ -0,0 +1,700 @@
...
+Apr 23 12:15:29 localhost.localdomain NetworkManager[704]: <info>  [1713874529.7945] device (eth0): state change: disconnected -> prepare (reason 'none', sys-iface-state: 'managed')
+Apr 23 12:15:29 localhost.localdomain NetworkManager[704]: <info>  [1713874529.7947] device (eth0): state change: prepare -> config (reason 'none', sys-iface-state: 'managed')
+Apr 23 12:15:29 localhost.localdomain NetworkManager[704]: <info>  [1713874529.7953] device (eth0): state change: config -> ip-config (reason 'none', sys-iface-state: 'managed')
+Apr 23 12:15:29 localhost.localdomain NetworkManager[704]: <info>  [1713874529.7964] dhcp4 (eth0): activation: beginning transaction (timeout in 90 seconds)
+Apr 23 12:15:33 localhost.localdomain NetworkManager[704]: <info>  [1713874533.1272] dhcp4 (eth0): state changed new lease, address=192.168.122.86
+
+localhost:~ # nmcli -f NAME,UUID,TYPE,DEVICE,FILENAME con show
+NAME              UUID                                  TYPE      DEVICE  FILENAME
+Wired Connection  300ed658-08d4-4281-9f8c-d1b8882d29b9  ethernet  eth0    /var/run/NetworkManager/system-connections/default_connection.nmconnection
+----
```
rdoxenham, line 700: Looks great. Maybe we should add an example for where we can do the "fallback" option we discussed via Slack @atanasdinov ? Just so we capture it in the docs for now as an example?

asciidoc/components/networking.adoc
hardys: @agracey the SLEMicro default is `net.ifnames=0`  -  however in the metal3 chart we default the configuration to `net.ifnames=1` and also document setting the corresponding setting in the ATIP docs.

In future it might be good to align on a single defaults, but there are pros/cons to either option hence my comment that this could be a good place to mention the option and what it does - I'm happy to push a follow-up with that addition though.

asciidoc/components/networking.adoc
```diff
@@ -0,0 +1,867 @@
...
+However, there is also the option to modify it via a custom script. Whilst this option is very flexible and is not MAC address dependant,
+its limitation stems from the fact that using it is much less convenient when bootstrapping multiple nodes with a single image.
+
+> NOTE: It is recommended to use the default network configuration via files describing the desired network states under the `/network` directory.
+Only opt for custom scripting when that behaviour is not applicable to your use case.
+
+We will build and provision an edge node using different configuration structure. Follow all steps starting from <<config_dir_creation>> up until <<default_network_definition>>.
+
+In this example, we will create a custom script which applies static configuration for the `eth0` interface on all provisioned nodes,
+as well as removing and disabling the automatically created wired connections by NetworkManager.
```
rdoxenham, line 715:
```suggestion
as well as removing and disabling the automatically created wired connections by NetworkManager. This is beneficial in situations where you want to make sure that every node in your cluster has an identical networking configuration, and as such you do not need to be concerned with the MAC address of each node prior to image creation.
```

PR #169, opened 2024-04-23
Pin the link to EIB upstream to the 1.0 release

jdob:
Our release docs should point into a release branch so that when main changes, users aren't shown things that don't exist in the 1.0 release.


atanasdinov (approved):

agracey (approved):
Good call

PR #171, opened 2024-04-24
Adding architecture drawing for Elemental quick start

agracey:


hardys:
Looks good to me - two small nits to consider:

- We should probably say k3s/RKE2 since this flow supports both and much of our focus elsewhere in the docs is around RKE2?
- We could perhaps make it clearer that the EIB image provides the registration URL to elemental-register, for example by extending the arrow (or adding a dotted additional arrow?) from "bootable image" to "elemental-register", then adding "registration URL" next to it?

agracey:
Merging as approval was given in slack

PR #224, opened 2024-04-26
ATIP docs align direct network provisioning naming

hardys:
Elsewhere we use the term "direct network provisioning" so align here for consistency and fix a few wording and format issues.


rdoxenham (approved):

alknopfler (approved):
Lgtm

PR #254, opened 2024-04-26
Metal3 quickstart updates

hardys:
Add details around building the base image and deploying the downstream clusters

### DIFF
--- a/asciidoc/quickstart/metal3.adoc
+++ b/asciidoc/quickstart/metal3.adoc
@@ -34,7 +34,7 @@ image::quickstart-metal3-architecture.png[]
 
 == Prerequisites
 
-There are also some specific constraints related to the downstream cluster server hardware and networking:
+There are some specific constraints related to the downstream cluster server hardware and networking:
 
 * Management cluster
  ** Must have network connectivity to the target server management/BMC API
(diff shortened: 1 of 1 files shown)


asciidoc/quickstart/metal3.adoc
```diff
@@ -191,6 +196,118 @@ clusterctl init --core "cluster-api:v1.6.2" --infrastructure "metal3:v1.6.0" --b
...
+}
+growfs /
+----
+
+[NOTE]
+====
+Add your own custom scripts to be executed during the provisioning process using the same approach.
+For more information, see <<quickstart-eib>>.
+
+The bug related to this workaround is https://bugzilla.suse.com/show_bug.cgi?id=1217430
```
alknopfler, line 292: Do we need to add the bug info explicitly to the docs?

asciidoc/quickstart/metal3.adoc
```diff
@@ -543,6 +725,53 @@ spec:
...
+
+
+=== Cluster deprovisioning
+
+The downstream cluster may be deprovisioned by deleting the resources applied in the creation steps above:
+
+[,bash]
+----
+% kubectl delete -f rke2-agent.yaml
+% kubectl delete -f rke2-control-plane.yaml
```
alknopfler, line 755: Something I noticed: if we're using the static IPs config secret, the finalizer is not well created so doing the -f to delete, the secret is deleted immediately and then the bmh is stuck on deprovisioning state , so based on my experience the best way to delete all resources is to delete the bmh object and then, during the process the secrets will be deleted

alknopfler (approved):
minor changes but in general looks good to me

asciidoc/quickstart/metal3.adoc
e-minguez: IMHO I think this is a good practice because users will have the background instead of being blind about that.

asciidoc/quickstart/metal3.adoc
hardys: Interesting - that's something we should perhaps raise upstream, I can add a note about that here but note the two deletes of the rke2 manifests in this example don't delete the BMH resources or associated secrets

PR #258, opened 2024-04-30
Add shasums for sle micro 5.5

e-minguez:
Adding shasums for Edge 3.0 as well.


e-minguez:
Shall we do this for every PR that we want to backport for 3.0 moving forward, right?

hardys:
> Shall we do this for every PR that we want to backport for 3.0 moving forward, right?

I discussed with @agracey and he suggested cherry-picks from main, but doing this via PRs perhaps enables easier tracking of the backports?

hardys (approved):

PR #261, opened 2024-05-02
Day 2 documentation

ipetrov117:
PR for Day 2 operations both on Rancher Management as well as Downstream Clusters.

**_Note:_** In order to have a functionally active documentation, all links that for the `suse-edge/fleet-examples` repository currently point to a fork. This will be fixed once the [Include file references for the "Day 2" documentation](https://github.com/suse-edge/fleet-examples/pull/10) PR in the `fleet-examples` is merged.

**_Note:_** I am aware that there is some documentation duplication throughout the SUC, OS, K8s docs. I wanted to have each section autonomous from one another. If you feel like this is too much, I am happy to try to extract common documentation out of the sections.

PR #287, opened 2024-06-17
day2: align with Rancher 2.8.4 version

hardys:
Followup to https://github.com/suse-edge/suse-edge.github.io/pull/261 so we can align with the 3.0.1 target version


ipetrov117 (approved):
Looks good.

PR #289, opened 2024-06-17
Day 2 documentation refactor for better compatibility with 3.0.1

ipetrov117:
- Helm chart upgrade section for management clusters has been made more generic and examples for upgrading both Rancher and Metal^3^ have been given. This is true for EIB and non-EIB deployed helm chart use-cases.
- Metal^3^ has been removed as an example for downstream cluster chart upgrades in favour of Longhorn to avoid confusion and for consistency.
- Minor changes, where made to some of the documentation text (suggested by @ranjinimn).


hardys (approved):

alknopfler (approved):
LGTM

PR #331, opened 2024-07-23
Adding translation files for weblate

rdoxenham:


PadraigDillon1 (approved):
Looks good to me. PO and POT files seem to have been created well.

PadraigDillon1:
In looking closer, the Admin guide seem to have separate PO/POT files for chapters or sections. To simplify the Weblate process it would be better if the guide was converted to a single PO/POT. This is how the SUMA team build their guides - for example, here is the Admin guide for SUMA converted to a single POT file even though there are many ascidoc source files:
https://github.com/uyuni-project/uyuni-docs/tree/master/l10n-weblate/administration
Otherwise it would be necessary to create a new Weblate project for each section/chatper for any new release rather than just one.

PR #334, opened 2024-08-01
fix some version issues with day2 scripts

alknopfler:
- fix some issues with the version in the day2 scripts usages


ranjinimn (approved):
Changes look good.

PR #368, opened 2024-09-11
Alignment with Rancher Turtles

hardys:
This aligns with the recently added suse-edge charts which enable management of CAPI dependencies via Rancher Turtles

Depends On:
https://github.com/suse-edge/charts/pull/150 (merged) and https://github.com/suse-edge/charts/pull/154


hardys:
Also see https://github.com/suse-edge/atip/pull/7

alknopfler (approved):
LGTM! Thanks Steven

hardys:
One thing to note is this currently refers to charts we've not yet published via registry.suse.com - I think we'll have that issue in other places, but if preferred I can update this to reference the suse-edge/charts upstream charts instead as a temporary measure

Kristian-ZH (approved):
lgtm

PR #376, opened 2024-09-16
Updating to latest versions for 3.1 release

dprodanov4:
Update docs and examples to our latest versions for 3.1 release.

PR #392, opened 2024-09-24
changed Longhorn chart location from longhorn.io to rancher.io

jdob:
Update the Longhorn charts location to use charts.rancher.io instead of charts.longhorn.io. See our (internal, sorry) Jira ticket EDGE-552.

Note: The upstream PR has landed (https://github.com/rancher/charts/tree/dev-v2.9/charts/longhorn/104.2.0%2Bup1.7.1) but it's not yet available in the registry, so I wasn't able to actually test these changes yet.

This PR does not include:
* Updates to the release notes. I added a comment to #376 that this is the version that will be available in that registry.
* I didn't touch `downstream-cluster-helm.adoc`. There's an entry in there on line 243 that references the old location. I can update this if it's safe; I didn't want to touch the day 2 stuff in case that was in progress and/or needs more information to update properly. Let me know and I'll update that reference too.

### DIFF
--- a/asciidoc/components/longhorn.adoc
+++ b/asciidoc/components/longhorn.adoc
@@ -55,11 +55,11 @@ For additional help installing open-iscsi, refer to the https://longhorn.io/docs
 There are several ways to install Longhorn on your Kubernetes clusters.
 This guide will follow through the Helm installation, however feel free to follow the https://longhorn.io/docs/1.6.1/deploy/install/[official documentation] if another approach is desired.
 
-. Add the Longhorn Helm repository:
+. Add the Rancher Charts Helm repository:
 +
 [,shell]
 ----
-helm repo add longhorn https://charts.longhorn.io
+helm repo add rancher-charts https://charts.rancher.io/
 ----
 +
 . Fetch the latest charts from the repository:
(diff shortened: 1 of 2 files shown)


atanasdinov (commented):
This is unfortunately not enough. While the upstream chart contains both CRDs and the other resources, the Rancher team splits that into two charts: Longhorn and [Longhorn CRD](https://github.com/rancher/charts/tree/dev-v2.9/charts/longhorn-crd). We need the second one too.

jdob:
I added the -crd chart and updated the air gap image list from https://raw.githubusercontent.com/longhorn/longhorn/v1.7.1/deploy/longhorn-images.txt

jdob:
@dprodanov4 this is going to conflict with #376 because that one updates the versions but still pulls from the longhorn registry, whereas this changes the registry location too (and adds in the separate -crd chart).

dprodanov4:
@jdob thank you for the heads-up will update mine now.

hardys:
> @dprodanov4 this is going to conflict with #376 because that one updates the versions but still pulls from the longhorn registry, whereas this changes the registry location too (and adds in the separate -crd chart).

Sorry missed this and merged #376 - this PR lgtm other than now needing a rebase 👍

hardys (approved):

hardys:
> @hardys conflicts resolved when you get a second to approve

Hmm, GH still says conflicts - I would suggest dropping the merge commit, then doing `git fetch -va && git rebase -i origin`

PR #413, opened 2024-09-30
Added a known issue (metal3 and elemental incompatibility on the same…

### LINKED ISSUE #412: Document limitations between metal3 and elemental for 3.0
3.1 seems to be ok but 3.0 cannot be deployed both elemental and metal3.

Ideally would like to add a link to somewhere where we track the issue (?)

e-minguez:
… cluster)

Closes #412


ranjinimn (approved):

hardys:
Thanks @e-minguez - indeed we should create a tracker issue for this, IIRC it's related to the CAPI configuration needed for the Metal3 flow, and with our move to Rancher Turtles it may possibly be resolved - although the Turtles installation currently does still require disabling the Rancher embedded CAPI controller and certain webhooks (which I think was related to the Elemental issue previously)

PR #434, opened 2024-10-01
Update NVIDIA guide for Edge 3.1

atanasdinov:


atanasdinov:
Needs someone with the appropriate hardware to verify this is working.

asciidoc/images/scc-packages-nvidia.png
rdoxenham, line 1: When you take a screenshot, press space-bar and you can select an entire window, and it gives you the transparent background rather than a border. Do you mind redoing this screenshot? Thanks!

asciidoc/integrations/nvidia-slemicro.adoc
```diff
@@ -36,19 +36,19 @@ Before we begin, it is important to recognize that besides the NVIDIA open-drive
 We recommend that you ensure that the driver version that you are selecting is compatible with your GPU and meets any CUDA requirements that you may have by checking:
 
 * The https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/[CUDA release notes]
-* The driver version that you plan on deploying has a matching version in the http://download.nvidia.com/suse/sle15sp5/x86_64/[NVIDIA SLE15-SP5 repository] and ensuring that you have equivalent package versions for the supporting components available 
+* The driver version that you plan on deploying has a matching version in the https://download.nvidia.com/suse/sle15sp6/x86_64/[NVIDIA SLE15-SP6 repository] and ensuring that you have equivalent package versions for the supporting components available
```
rdoxenham, line 39: Do we know whether SP6 binaries are the right ones to use here? I suspect that ideally we need SLFO-based binaries instead. I'll speak with the NVIDIA team.

manuelbuil:
This PR gave me good information to solve a problem I was having with GPU Tesla M30 (g3s.xlarge flavor in AWS) and I feel in debt with your guy so  I just wanted to mention my problems as it might be helpful to include in the docs. If not, just ignore.
When GPUs are a bit old, like Tesla M30, nvidia-open-drivers do not work, in that case only proprietary kernel drivers work (maybe you only support open kernel drivers). Moreover, when using SLE15, the proprietary drivers in `https://download.nvidia.com/suse/sle15sp6/` do not work. But the proprietary drivers in `https://developer.download.nvidia.com/compute/cuda/repos/sles15/x86_64/` work. I don't understand what repo is recommended by NVIDIA. Both provide drivers but different versions of those drivers, which is confusing

rdoxenham (approved):

PR #444, opened 2024-10-02
Update EIB quickstart

atanasdinov:


asciidoc/quickstart/eib.adoc
```diff
@@ -17,9 +17,9 @@ For more information, read the <<components-eib,Edge Image Builder Introduction>
 
 == Prerequisites
 
-* An x86_64 physical host (or virtual machine) running SLES 15 SP5, openSUSE Leap 15.5, or openSUSE Tumbleweed.
+* An x86_64 physical host (or virtual machine) running SLES 15 SP6, openSUSE Leap 15.6, or openSUSE Tumbleweed.
 * An available container runtime (e.g. Podman)
-* A downloaded copy of the latest SLE Micro 5.5 SelfInstall "GM2" ISO image found https://www.suse.com/download/sle-micro/[here].
+* A downloaded copy of the latest SLE Micro 6.0 SelfInstall ISO image found https://www.suse.com/download/sle-micro/[here].
```
hardys, line 22: We'll have to reinstate the GM2 here when published...

hardys (approved):

asciidoc/quickstart/eib.adoc
atanasdinov: Should I revert that part? We'll be adjusting the docs with the new image anyway across the board.

hardys: I think it's fine to leave it as-is for now, we'll just have to adjust in a follow-up when the 6.0 GM2 is published

asciidoc/quickstart/eib.adoc
```diff
@@ -42,7 +42,7 @@ export CONFIG_DIR=$HOME/eib
 mkdir -p $CONFIG_DIR/base-images
 ----
 
-In the previous step we created a "base-images" directory that will host the SLE Micro 5.5 input image, let's ensure that the downloaded image is copied over to the configuration directory:
+In the previous step we created a "base-images" directory that will host the SLE Micro 6.0 input image, let's ensure that the downloaded image is copied over to the configuration directory:
```
jdob, line 45: I think we should be using "SL Micro" instead of "SLE", even though I know there's a bit of confusion. But that's tangential to this PR; I'll make a task to revisit in the future.

jdob (approved):

asciidoc/quickstart/eib.adoc
atanasdinov: Well, from my understanding even that is not the correct name, but yeah, let's resolve it in a follow-up.

PR #424, opened 2024-09-30
Adjust airgapped EIB guide

atanasdinov:

PR #447, opened 2024-10-02
atip-management-cluster: update versions

hardys:
Align target versions with 3.1 release and airgap images from #424 and #448

### DIFF
--- a/asciidoc/product/atip-management-cluster.adoc
+++ b/asciidoc/product/atip-management-cluster.adoc
@@ -163,19 +163,19 @@ kubernetes:
     charts:
       - name: cert-manager
         repositoryName: jetstack
-        version: 1.14.2
+        version: 1.15.3
         targetNamespace: cert-manager
         valuesFile: certmanager.yaml
         createNamespace: true
         installationNamespace: kube-system
       - name: longhorn-crd
-        version: 103.3.0+up1.6.1
+        version: 104.2.0+up1.7.1
         repositoryName: rancher-charts
         targetNamespace: longhorn-system
         createNamespace: true
         installationNamespace: kube-system
       - name: longhorn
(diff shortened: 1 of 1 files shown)


asciidoc/product/atip-management-cluster.adoc
```diff
@@ -1159,11 +1159,11 @@ embeddedArtifactRegistry:
     - name: registry.rancher.com/rancher/mirrored-longhornio-longhorn-share-manager:v1.6.1
     - name: registry.rancher.com/rancher/mirrored-longhornio-longhorn-ui:v1.6.1
     - name: registry.rancher.com/rancher/mirrored-longhornio-support-bundle-kit:v0.0.36
-    - name: registry.suse.com/edge/3.1/cluster-api-provider-rke2-bootstrap:v0.4.1
```
hardys, line 1162: @atanasdinov I need to align this with #424 but that doesn't include the longhorn images, can you share any details re how you generated that list please?

atanasdinov: Initially, I used the images coming from https://raw.githubusercontent.com/longhorn/longhorn/v1.7.1/deploy/longhorn-images.txt. I built an image with EIB and provisioned a node in an airgapped environment. I quickly realised that this will not work with Longhorn installed through the Rancher chart since the latter would use the mirrored images instead which is why I manually replaced those and reran the aforementioned procedure.

hardys: Ok thanks, I updated to align with https://github.com/suse-edge/suse-edge.github.io/pull/424 and https://github.com/suse-edge/suse-edge.github.io/pull/448 - PTAL, thanks!

atanasdinov (approved):

PR #448, opened 2024-10-02
Switch Longhorn images source

atanasdinov:


asciidoc/guides/air-gapped-eib-deployments.adoc
```diff
@@ -566,21 +566,21 @@ kubernetes:
...
+    - name: registry.suse.com/rancher/mirrored-longhornio-livenessprobe:v2.14.0
+    - name: registry.suse.com/rancher/mirrored-longhornio-openshift-origin-oauth-proxy:4.15
+    - name: registry.suse.com/rancher/mirrored-longhornio-backing-image-manager:v1.7.1
+    - name: registry.suse.com/rancher/mirrored-longhornio-longhorn-engine:v1.7.1
+    - name: registry.suse.com/rancher/mirrored-longhornio-longhorn-instance-manager:v1.7.1
+    - name: registry.suse.com/rancher/mirrored-longhornio-longhorn-manager:v1.7.1
+    - name: registry.suse.com/rancher/mirrored-longhornio-longhorn-share-manager:v1.7.1
+    - name: registry.suse.com/rancher/mirrored-longhornio-longhorn-ui:v1.7.1
+    - name: registry.suse.com/rancher/mirrored-longhornio-support-bundle-kit:v0.0.42
     - name: longhornio/longhorn-cli:v1.7.1
```
atanasdinov, line 583: I'm wondering if we should drop this, as it is an image that Rancher doesn't mirror and a deployment fully works without it but I guess it doesn't really get in the way anyway.

hardys (approved):
lgtm - started a discussion re longhorn-cli and we can deal with that as a follow-up if needed

Issue #389, opened 2024-09-24
[doc] Issue in "Create downstream clusters" - Duplicate label


peteingithub:
Create downstream clusters:

https://documentation.suse.com/suse-edge/3.0/html/edge/quickstart-elemental.html#create-downstream-clusters

```
INET=`ip addr show dev eth0 | grep "inet\ "`
elemental-register **--label "network=$INET" \
   --label "network=$INET"** /oem/registration
 ```

PR #449, opened 2024-10-03
elemental-quickstart fixes

hardys:
Fixes: #389


atanasdinov (dismissed):

jdob (dismissed):

rdoxenham:
I don't actually think this makes sense at all given that we're not using the `/oem/` path, as that's for Elemental Teal

hardys:
> I don't actually think this makes sense at all given that we're not using the `/oem/` path, as that's for Elemental Teal

I did wonder about that - do we just remove the `/oem/registration` or is the labelling itself not relevant?  (sorry not had time to dig into the details yet)

hardys:
> Agreed, I think the whole section can be removed.

Ok now removed, and I noticed the Rancher version needed updating

atanasdinov (approved):

rdoxenham (approved):

jdob (approved):

PR #456, opened 2024-10-09
Add 3.0 to 3.1 migration documentation

ipetrov117:
Adds migration documentation to our "Day 2" docs that specifies the steps users should take in order to migrate their 3.0.X Edge environments to Edge 3.1.


asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,1004 @@
...
+
+This section covers how to migrate a `management` cluster from `Edge 3.0` to `Edge 3.1`.
+
+`Management` cluster component should be migrated in the following order:
+
+. <<day2-migration-mgmt-os, Operating System (OS)>>
+
+. <<day2-migration-mgmt-rke2, RKE2>>
+
+. <<day2-migration-mgmt-helm, SUSE Helm charts>>
```
atanasdinov, line 30: I'm wondering if there's a specific term we should use for those. There are at least few Helm charts that we just copy as is from other sources so claiming they're "SUSE" might not be appropriated. @hardys thoughts?

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,1004 @@
...
+
+. Check the <<day2-migration-mgmt-helm-limitations, Known Limitations>> section if there are any additional steps, or changes that need to be done for the charts.
+
+. link:https://helm.sh/docs/helm/helm_upgrade/[Upgrade] the helm chart to the desired version:
+
+** For non air-gapped setups:
++
+[,bash]
+----
+# For Helm repo based charts
```
atanasdinov, line 693:
```suggestion
# For charts hosted in HTTP repositories
```

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,1004 @@
...
+helm upgrade <chart_name> <chart_repo>/<chart_name> --version <version> --values <chart_name>-values.yaml -n <chart_namespace>
+
+# For OCI based charts
```
atanasdinov, line 696:
```suggestion
# For charts hosted in OCI registries
```

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,1004 @@
...
+** For air-gapped setups:
+
+*** On a machine with access to the internet, pull the desired chart version:
++
+[,bash]
+----
+# For Helm repo based charts
+helm pull <chart_repo_name>/<chart_name> --version=X.Y.Z
+
+# For OCI based charts
```
atanasdinov, line 709:
```suggestion
# For charts hosted in OCI registries
```

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,1004 @@
...
+helm upgrade <chart_name> oci://<oci_registry_url>/<chart_name> --namespace <chart_namespace> --values <chart_name>-values.yaml --version=X.Y.Z
+----
+
+** For air-gapped setups:
+
+*** On a machine with access to the internet, pull the desired chart version:
++
+[,bash]
+----
+# For Helm repo based charts
```
atanasdinov, line 706:
```suggestion
# For charts hosted in HTTP repositories
```

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,1004 @@
...
+[,bash]
+----
+scp <chart>.tgz <machine-address>:<filesystem-path>
+----
+
+*** Upgrade the chart:
++
+[,bash]
+----
+helm upgrade <chart_name> <chart>.tgz --values <chart_name>-values.yaml -n <chart_namespace>
```
atanasdinov, line 724: Can we assume that `helm` will be available on the cluster node?

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,1004 @@
...
+[,bash]
+----
+helm repo add rancher-prime https://charts.rancher.com/server-charts/prime
+----
+
+** The `Metal^3^` is deployed from the `registry.suse.com` OCI registry in the `metal3-system` namespace.
+
+_Steps:_
+
+. Locate the desired versions for `Rancher` and `Metal^3^` from the <<release-notes, release notes>>. For `Edge 3.1`, these versions would be `2.9.1` for Rancher and `0.8.1` for Metal^3^.
```
atanasdinov, line 758:
```suggestion
. Locate the desired versions for `Rancher` and `Metal^3^` from the <<release-notes, release notes>>. For `Edge 3.1.0`, these versions would be `2.9.1` for Rancher and `0.8.1` for Metal^3^.
```

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,1004 @@
...
+== Downstream clusters
+
+This section covers how to migrate your `Edge 3.0.X` downstream clusters to `Edge 3.1`.
+
+[#day2-migration-downstream-prerequisites]
+=== Prerequisites
+
+This section covers any prerequisite steps that users should go through before beginning the migration process.
+
+==== Charts deployed through the Edge Image Builder
```
atanasdinov, line 807: Can we unify this across the document - there are "EIB", "non-EIB", "Edge Image Builder" references and it'd be nice if we can reuse the same wording.

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,1004 @@
...
+kubectl get bundle eib-chart-migration-prep -n fleet-default
+NAME                       BUNDLEDEPLOYMENTS-READY   STATUS
+eib-chart-migration-prep   1/1                 
+----
+
+. By creating the above `Bundle` on your `management` cluster, Fleet will deploy a `SUC Plan` (called `eib-chart-migration-prep`) on each downstream cluster that matches the `targets` specified in the `targets.yaml` file. To monitor the lifecycle of this plan, refer to <<components-system-upgrade-controller-monitor-plans, "Monitoring System Upgrade Controller Plans">>.
+
+=== Migration steps
+
+After executing the <<day2-migration-downstream-prerequisites, prerequisite>> steps, you can proceed to follow the <<day2-downstream-clusters, downstream cluster>> upgrade documentation for the `Edge 3.1` release.
```
atanasdinov, line 1004: Do we need this section?

atanasdinov (commented):
LGTM! Raised some points for discussion but happy to approve after we align on those.

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,1004 @@
...
+[,bash]
+----
+kubectl patch helmchart rancher -n <namespace> --type='merge' -p '{"spec":{"set":{"ingress.ingressClassName":"nginx"}}}'
+----
+
+* If `Rancher` was deployed through Helm, add the `--set ingress.ingressClassName=nginx` flag to your upgrade command:
++
+[,bash]
+----
+helm upgrade rancher <chart_repo>/rancher --version <version> --set ingress.ingressClassName=nginx --values <chart_name>-values.yaml -n <chart_namespace>
```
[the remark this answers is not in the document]
ipetrov117: Well I can't say I agree, the fact that we need to do this is a limitation in itself, so IMO it should be under `Known Limitations`. 

I agree, however, that we do not need to have the command here. How about just specifying that when users upgrade rancher they need to add the `--set ingress.ingressClassName=nginx` option to the helm upgrade command?

asciidoc/day2/migration.adoc
ipetrov117: On the management cluster, IMO yes. It is reasonable to assume that even for air-gapped systems users would have `helm` setup on their management cluster. IMO, generally this cluster should have such tooling even on air-gapped systems.

We can add a note saying that this documentation expects helm to be available. WDYT?

asciidoc/day2/migration.adoc
ipetrov117: IMO, yes. This marks the logical next step to the downstream cluster update mechanism. Without it users will not be sure where to proceed once the `prerequisite` steps have concluded.

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,1004 @@
...
+:warning-caption: :warning:
+endif::[]
+:toc: preamble
+
+This section covers how to migrate existing `Edge 3.0` <<day2-migration-mgmt,management>> and <<day2-migration-downstream, downstream>> clusters to `Edge 3.1`.
+
+For a list of `Edge 3.1` component versions, refer to the <<release-notes, release notes>>.
+
+[#day2-migration-mgmt]
+== Management cluster
```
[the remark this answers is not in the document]
ipetrov117: It will look like this for the root "Day 2" section:
Screenshot 2024-10-09 at 14 33 49

And like this for the "Edge 3.1 migration" section:
Screenshot 2024-10-09 at 14 33 59

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,1004 @@
...
+:imagesdir: ../images/
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:important-caption: :heavy_exclamation_mark:
+:caution-caption: :fire:
+:warning-caption: :warning:
+endif::[]
+:toc: preamble
+
+This section covers how to migrate existing `Edge 3.0` <<day2-migration-mgmt,management>> and <<day2-migration-downstream, downstream>> clusters to `Edge 3.1`.
```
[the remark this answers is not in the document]
ipetrov117: IMO it should reside only on 3.1. Regarding whether to put a link to 3.0, I am not sure where this link would reside. 

I am not sure of the value to have this in 3.0 to be honest, mainly because, as a user I  would pick up the 3.1 documentation directly and go to the `Edge 3.1 migration` documentation. That said, if you think that we need this in 3.0, I am not opposed to adding a link to it.

asciidoc/day2/migration.adoc
ipetrov117: I kept the `SUSE` helm charts mainly because this was the previous "Day 2" naming. 

I agree that they are not specifically charts of SUSE, but they are Edge charts and that was the distinction I was trying to make - that this process is for Edge charts only.

How about naming this "Edge Helm charts" instead? WDYT?

asciidoc/day2/migration.adoc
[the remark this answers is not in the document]
atanasdinov: Yeah, this is exactly what I meant. We need the section and we need to use it to describe what has to be done for whatever reason, but the upgrade step doesn't necessarily belong here. I think specifying the parameter is more than enough.

asciidoc/day2/migration.adoc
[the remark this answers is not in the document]
atanasdinov: Thanks! My concern is the actual page (in the second screenshot).

I think it'd read much better overall if the ToC to the right contains the different sub-steps instead of only distinguishing between management and downstream clusters. Perhaps that's impossible without splitting this document into separate ones for the migration of each. Then each one would be easier to navigate.

This might be a very subjective take, though. WDYT? Also bringing in @hardys and @ranjinimn in this discussion.

asciidoc/day2/migration.adoc
atanasdinov: To be honest, I'd rather leave this decision to PM. My only worry is that we might run into some issues with legal if we claim "ownership" using SUSE's name but that might also be just my overthinking.

asciidoc/day2/migration.adoc
atanasdinov: I'm fine with keeping this as is. It was purely a question so if the assumption is that Helm would be available, that's fine.

asciidoc/day2/migration.adoc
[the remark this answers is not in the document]
ipetrov117: Yea, if they are in separate files that are referred here, then the ToC will be a bit more descriptive, I agree.

It is not really a problem for me to do this. I handle it after all the other changes.

asciidoc/day2/migration.adoc
atanasdinov: This might be somewhat related to https://github.com/suse-edge/suse-edge.github.io/pull/456#discussion_r1793376104 and how we lay out the structure itself.

asciidoc/day2/migration.adoc
ipetrov117: I like `Edge Helm charts` and we can then have a description in the section distinguishing what we actually mean by `Edge Helm charts` -> charts that make up the Edge platform. IMO this is descriptive enough without landing into legal trouble. But I'll let @hardys pitch in.

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,1004 @@
...
+To avoid any unforeseen problems, migrate the clusters' `control-plane` nodes first and the `worker` nodes second.
+====
+
+==== Prerequisites
+
+* `SCC registered nodes` -  ensure your cluster nodes' OS are registered with a subscription key that supports the operating system version specified in the `Edge 3.1` <<release-notes,release>>.
+
+_Air-gapped:_
+
+* `Mirror SUSE RPM repositories` - RPM repositories related to the operating system that is specified in the `Edge 3.1` <<release-notes,release>> should be locally mirrored, so that `transactional-update` has access to them. This can be achieved by using either link:https://github.com/SUSE/rmt[RMT] or link:https://documentation.suse.com/suma/4.3/en/suse-manager/index.html[SUMA].
```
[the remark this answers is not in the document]
e-minguez: Also, SUMA 5.0 https://documentation.suse.com/suma/5.0/en/suse-manager/index.html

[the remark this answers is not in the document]
ipetrov117: Yea that is the one, although it is frustrating how it is for SLES and not exactly for SL Micro. I know they share the same kernel, but..

asciidoc/day2/migration.adoc
ipetrov117: Removed the `Edge Image Builder`, now we refer it only as `EIB` or for use-cases that are not EIB related - `non-EIB`.

asciidoc/day2/migration.adoc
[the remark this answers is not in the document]
ipetrov117: Sadly I was not able to achieve this, even after I split the `management` and `downstream` clusters in separate files that were referred in the `migration.adoc` file, it still shows the above ToC. Mainly because it merges the 2 adocs into one I think.. 

Also I tried playing with [:toclevels:](https://docs.asciidoctor.org/asciidoc/latest/toc/levels/), but it still did not work.

@ranjinimn, can you give some advice as to what I am doing wrong here. Generally we want for the deeper headers to be visible inside of the ToC, so that it can be more readable for the user.

asciidoc/day2/migration.adoc
ipetrov117: I changed the general naming for the charts to `Edge Helm charts`. @hardys, lemme know if you do not agree with the naming.

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,993 @@
...
+:imagesdir: ../images/
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:important-caption: :heavy_exclamation_mark:
+:caution-caption: :fire:
+:warning-caption: :warning:
+endif::[]
+:toc: preamble
+
+This section offers migration guidelines for existing `Edge 3.0` *management* and *downstream* clusters to `Edge 3.1.0`.
```
e-minguez, line 15:
```suggestion
This section offers migration guidelines for existing `Edge 3.0` (including minor releases such as 3.0.1 and 3.0.2) *management* and *downstream* clusters to `Edge 3.1.0`.
```

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,993 @@
...
+
+For a list of `Edge 3.1.0` component versions, refer to the <<release-notes, release notes>>.
+
+[#day2-migration-mgmt]
+== Management cluster
+
+This section covers how to migrate a `management` cluster from `Edge 3.0` to `Edge 3.1.0`.
+
+`Management` cluster component should be migrated in the following order:
```
e-minguez, line 24:
```suggestion
`Management` cluster components should be migrated in the following order:
```

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,993 @@
...
+----
+transactional-update --continue migration
+----
++
+[NOTE]
+====
+The `--continue` option is used here in order to reuse the previous snapshot without having to reboot the system.
+====
+
+** If your subscription key supports the `SUSE Linux Micro 6.0` version, you will be prompted with something similar to:
```
e-minguez, line 100: I think this is more a prerequisite

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,993 @@
...
++
+image::day2-migration-os-migration-prompt.png[]
++
+Select the `number` that corresponds to `SUSE Linux Micro 6.0 <arch>`.
```
e-minguez, line 104: I'd say to add a note saying Edge 3.1 is SUSE Linux Micro 6.0 ONLY

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,993 @@
...
+
+. After a successful `transactional-update` run, for the changes to take effect on the system you would need to reboot:
++
+[,bash]
+----
+reboot
+----
+
+. Validate that the operating system has been migrated to `SUSE Linux Micro 6.0`:
```
e-minguez, line 113:
```suggestion
. After the host has been rebooted, validate that the operating system has been migrated to `SUSE Linux Micro 6.0`:
```

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,993 @@
...
+----
+kubectl uncordon <node_name>
+----
+
+[#day2-migration-mgmt-helm]
+=== Edge Helm charts
+
+[NOTE]
+====
+This section assumes you have installed `helm` on your system. For `helm` installation instructions, check link:https://helm.sh/docs/intro/install[here].
```
e-minguez, line 227: I'd say to add something like 'you need a valid kubeconfig file pointing to your cluster' as well.

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,993 @@
...
+----
+====
+
+Before upgrading `Rancher`, make sure to execute the following command:
+
+* If `Rancher` was deployed through <<components-eib, EIB>>:
++
+[,bash]
+----
+kubectl patch helmchart rancher -n <namespace> --type='merge' -p '{"spec":{"set":{"ingress.ingressClassName":"nginx"}}}'
```
e-minguez, line 269:
```suggestion
kubectl patch helmchart rancher -n cattle-system --type='merge' -p '{"spec":{"set":{"ingress.ingressClassName":"nginx"}}}'
```

If it was deployed via EIB we know the namespace already, right?

ipetrov117: No we don't, users can deploy the `HelmChart` resource to any namespace they wish.

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,993 @@
...
+----
+
+* If `Rancher` was deployed through Helm, add the `--set ingress.ingressClassName=nginx` flag to your link:https://helm.sh/docs/helm/helm_upgrade/[upgrade] command. For a full example of how to utilize this option, see the following <<day2-migration-mgmt-helm-non-eib-example, example>>.
+
+[#day2-migration-mgmt-helm-capi]
+==== Cluster API controllers migration
+
+From `Edge 3.1.0`, Cluster API (CAPI) controllers on a Metal^3^ management cluster are managed via the `Rancher Turtles` Helm chart.
```
e-minguez, line 277:
```suggestion
From `Edge 3.1.0`, Cluster API (CAPI) controllers on a Metal^3^ management cluster are managed via link:https://turtles.docs.rancher.com[Rancher Turtles].
```

asciidoc/day2/migration.adoc
ipetrov117: There is an `installationNamespace` configuration in the EIB definition file, that can specify this. Furthermore, if it is missing the default namespace is used.

asciidoc/day2/migration.adoc
ipetrov117: We have it under prerequisites [here](https://github.com/ipetrov117/suse-edge.github.io/blob/day2-migration-docs/asciidoc/day2/migration.adoc?plain=1#L48), this section is used to additionally highlight the need for the correct key.

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,993 @@
...
+. Get all EIB deployed manifests:
++
+[NOTE]
+====
+All EIB deployed manifests will have the `"edge.suse.com/source: edge-image-builder"` label.
+====
++
+[,bash]
+----
+grep -rl 'edge.suse.com/source: edge-image-builder'
```
e-minguez, line 423:
```suggestion
grep -lrIZ 'edge.suse.com/source: edge-image-builder' /var/lib/rancher/rke2/server/manifests | xargs -0 rm -f --
```

e-minguez: Or even better, copy them elsewhere just in case 🗡️

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,993 @@
...
+[WARNING]
+====
+If this is not done, any chart upgrade will be reverted by the RKE2/K3s process upon restart of the process or the operating system.
+====
+
+On downstream clusters, the removal of the EIB created chart manifest files is handled by a Fleet called link:https://github.com/suse-edge/fleet-examples/tree/main/fleets/day2/system-upgrade-controller-plans/eib-charts-migration-prep[eib-charts-migration-prep] located in the link:https://github.com/suse-edge/fleet-examples.git[suse-edge/fleet-examples] repository.
+
+[WARNING]
+====
+Using the `eib-charts-migration-prep` Fleet from `main` is *not* advised. The fleet should *always* be used from a valid Edge link:https://github.com/suse-edge/fleet-examples/releases[release] tag.
```
e-minguez, line 811:
```suggestion
Using the `eib-charts-migration-prep` Fleet file from the `main` branch is *not* advised. The Fleet file should *always* be used from a valid Edge link:https://github.com/suse-edge/fleet-examples/releases[release] tag.
```

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,993 @@
...
+====
+
+[IMPORTANT]
+====
+This process requires that SUC is already deployed. For installation details, refer to <<components-system-upgrade-controller-install, "Installing the System Upgrade Controller">>.
```
e-minguez, line 816:
```suggestion
This process requires that system-upgrade-controller (SUC) is already deployed. For installation details, refer to <<components-system-upgrade-controller-install, "Installing the System Upgrade Controller">>.
```

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,993 @@
...
+====
+
+Once created, the `eib-charts-migration-prep` Fleet ships a <<components-system-upgrade-controller,SUC>> Plan that contains a script that will do the following:
+
+. Determine whether the current node it is running on is an `initializer` node.
```
e-minguez, line 821:
```suggestion
. Determine whether the current node it is running on is an `initializer` node. If it is not, it won't do anything.
```

Is that right?

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,993 @@
...
+----
+cat > targets.yaml <<EOF
+targets:
+- clusterSelector: CHANGEME
+EOF
+----
++
+For information on how to map to downstream clusters, see link:https://fleet.rancher.io/gitrepo-targets[Mapping to Downstream Clusters].
+
+. Proceed to building the Bundle:
```
e-minguez, line 950:
```suggestion
. Proceed to build the Bundle:
```

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,993 @@
...
+
+** For Mac users, there is a link:https://formulae.brew.sh/formula/fleet-cli[fleet-cli] Homebrew Formulae.
+
+** For Linux users, the binaries are present as *assets* to each Fleet link:https://github.com/rancher/fleet/releases[release].
+
+*** Linux AMD:
++
+[,bash]
+----
+curl -L -o fleet-cli https://github.com/rancher/fleet/releases/download/<FLEET_VERSION>/fleet-linux-amd64
```
e-minguez, line 914:
```suggestion
curl -L -o fleet-cli https://github.com/rancher/fleet/releases/download/<FLEET_VERSION>/fleet-linux-amd64
sudo mkdir -p /usr/local/bin
sudo mv ./fleet-cli /usr/local/bin/fleet-cli
sudo chmod 755 /usr/local/bin/fleet-cli
```

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,993 @@
...
+----
+
+*** Linux ARM:
++
+[,bash]
+----
+curl -L -o fleet-cli https://github.com/rancher/fleet/releases/download/<FLEET_VERSION>/fleet-linux-arm64
```
e-minguez, line 921:
```suggestion
curl -L -o fleet-cli https://github.com/rancher/fleet/releases/download/<FLEET_VERSION>/fleet-linux-arm64
sudo mkdir -p /usr/local/bin
sudo mv ./fleet-cli /usr/local/bin/fleet-cli
sudo chmod 755 /usr/local/bin/fleet-cli
```

asciidoc/day2/migration.adoc
ipetrov117: Yep, I added both backup (copy) and removal instructions.

asciidoc/day2/migration.adoc
ipetrov117: Section where separated to:

1. Binary retrieval
2. Binary move to `/usr/local/bin

asciidoc/day2/migration.adoc
ipetrov117: Section where separated to:

1. Binary retrieval
2. Binary move to `/usr/local/bin

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,993 @@
...
+
+. <<day2-migration-mgmt-os, Operating System (OS)>>
+
+. <<day2-migration-mgmt-rke2, RKE2>>
+
+. <<day2-migration-mgmt-helm, Edge Helm charts>>
+
+
+[#day2-migration-mgmt-os]
+=== Operating System (OS)
```
Kristian-ZH, line 34: Why we do not suggest here the SUC plans? So many manual steps which are already handled in our SUC approach

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,993 @@
...
+For a full list of the options for the `drain` command, see link:https://kubernetes.io/docs/reference/kubectl/generated/kubectl_drain/[kubectl drain].
+
+. Use the RKE2 installation script to install the correct `Edge 3.1.0` compatible RKE2 version:
++
+[,bash]
+----
+curl -sfL https://get.rke2.io | INSTALL_RKE2_VERSION=v1.30.3+rke2r1 sh -
+----
+
+. Restart the `rke2` process:
```
Kristian-ZH, line 202: Same as above. All of those things are already handled with the SUC plans

ipetrov117: `3.0` systems do not have SUC deployed by default. Furthermore `management` clusters where small enough that it was agreed that manual steps would be preferred over the SUC approach. We decided on this when we were writing the `3.0` documentation. That is why I decided to keep the manual steps.

asciidoc/day2/migration.adoc
ipetrov117: `3.0` systems do not have SUC deployed by default. Furthermore `management` clusters where small enough that it was agreed that manual steps would be preferred over the SUC approach. We decided on this when we were writing the `3.0` documentation. That is why I decided to keep the manual steps.

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,993 @@
...
+[#day2-migration-mgmt-helm-capi]
+==== Cluster API controllers migration
+
+From `Edge 3.1.0`, Cluster API (CAPI) controllers on a Metal^3^ management cluster are managed via the `Rancher Turtles` Helm chart.
+
+To migrate the CAPI controllers versions to `Edge 3.1.0` compatible versions, install the `Rancher Turtles` chart:
+
+[,bash]
+----
+helm install rancher-turtles oci://registry.suse.com/edge/3.1/rancher-turtles-chart --version 0.3.2 --namespace rancher-turtles-system --create-namespace
```
Kristian-ZH, line 288: I would suggest installing the `turtles` via `HelmChart` resource. In this way, it will be the same as the other components installed by EIB, and also, the upgraded cluster will look exactly the same way as EIB provisioned 3.1 cluster

ipetrov117: Isn't the `registry.suse.com` the source of truth when it comes to our helm charts? AFAIK we prefer it over referring the `charts` repository.

ipetrov117: Why would I need to try to match it to what EIB deploys? Aren't we trying to give the users the easiest command to install the helm chart? 

IMO doing it through helm is the better approach. Mainly because it is easier and the `upgrade-controller` does not care if the chart is deployed through a `HelmChart` resource or not, it will still upgrade it in the future.

asciidoc/day2/migration.adoc
Kristian-ZH: this is a worse option to me...  We already have those Plans. Now everything that we need to suggest is.
1. Deploy SUC (if the cluster is MNGM)
2. Apply the plans

Why should we support 2 different flows when they are similar and we already have a solution?

Anyway... Ignore this comment then

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,993 @@
...
+[,bash]
+----
+helm pull oci://<chart_oci_url> --version=X.Y.Z
+----
+
+. Encode the pulled chart archive:
++
+[source,bash]
+----
+base64 -w 0 <chart_name>-X.Y.Z.tgz  > <chart_name>-X.Y.Z.txt
```
Kristian-ZH, line 468: And here can't we reuse our scripts?

ipetrov117: Same answer as above.

ranjinimn (commented):
@ipetrov117 I have a few suggestions. 
Apart from that, I will need to review the local PDF version of the document to let you know about the sections and sub sections. Can you please upload it here? Else, I will generate a PDF and check if it the sections and sub sections look okay.

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,993 @@
...
+:toc: preamble
+
+This section offers migration guidelines for existing `Edge 3.0` *management* and *downstream* clusters to `Edge 3.1.0`.
+
+For a list of `Edge 3.1.0` component versions, refer to the <<release-notes, release notes>>.
+
+[#day2-migration-mgmt]
+== Management cluster
+
+This section covers how to migrate a `management` cluster from `Edge 3.0` to `Edge 3.1.0`.
```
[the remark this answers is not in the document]
ipetrov117: We are migrating a `3.0` cluster to the Edge `3.1.0` release. I am not sure what the best versioning convention should be here. Lemme know if I should change this to `3.1` instead.

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,1022 @@
...
+For a full list of the options for the `drain` command, see link:https://kubernetes.io/docs/reference/kubectl/generated/kubectl_drain/[kubectl drain].
+
+. Before a migration, you need to ensure that packages on your current OS are updated. To do this, execute:
++
+[,bash]
+----
+transactional-update
+----
++
+The above command executes link:https://en.opensuse.org/SDB:Zypper_usage#Updating_packages[zypper up] to update the OS packages. For more information on `transactional-update`, see the link:https://kubic.opensuse.org/documentation/transactional-update-guide/transactional-update.html[transactional-update guide].
```
e-minguez, line 86:
```suggestion
The above command executes link:https://en.opensuse.org/SDB:Zypper_usage#Updating_packages[zypper up] to update the OS packages. For more information on `transactional-update`, see the link:https://documentation.suse.com/smart/systems-management/html/Micro-transactional-updates/index.html[transactional-update guide].
```

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,1022 @@
...
+[NOTE]
+====
+In case something failed with the migration, you can rollback to the last working snapshot using:
+
+[,bash]
+----
+transactional-update rollback last
+----
+
+You would need to reboot your system for the `rollback` to take effect.
```
e-minguez, line 150:
```suggestion
You would need to reboot your system for the `rollback` to take effect. See link:https://documentation.suse.com/smart/systems-management/html/Micro-transactional-updates/index.html#tr-up-rollback[the official `transactional-update` documentation] for more information about the rollback procedure.
```

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,1022 @@
...
+----
+
+[#day2-migration-mgmt-rke2]
+=== RKE2
+
+[IMPORTANT]
+====
+The below steps should be done for each node of the `management` cluster.
+
+To avoid any unforeseen problems, migrate the clusters' `control-plane` nodes first and the `worker` nodes second.
```
e-minguez, line 167:
```suggestion
As the link:https://docs.rke2.io/upgrade/manual_upgrade[RKE2 documentation] explains, the upgrade procedure requires to upgrade the clusters' `control-plane` nodes one at a time and once all have been upgraded, the `agent` nodes.
```

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,1022 @@
...
+----
+kubectl uncordon <node_name>
+----
+
+[#day2-migration-mgmt-helm]
+=== Edge Helm charts
+
+[NOTE]
+====
+This section assumes you have installed `helm` on your system and you have a valid `kubeconfig` pointing to the desired cluster. For `helm` installation instructions, check link:https://helm.sh/docs/intro/install[here].
```
e-minguez, line 232:
```suggestion
This section assumes you have installed `helm` on your system and you have a valid `kubeconfig` pointing to the desired cluster. For `helm` installation instructions, check the link:https://helm.sh/docs/intro/install[Installing Helm] guide.
```

asciidoc/day2/migration.adoc
e-minguez: I agree with Kristian about this, specially to avoid differences between upgraded clusters and fresh ones. A user would be confused if he does `kubectl get helmcharts` and the output is different.
My 2 cents.

asciidoc/day2/migration.adoc
```diff
@@ -0,0 +1,1022 @@
...
+
+. On a machine with internet access pull the `rancher-turtles-airgap-resources` and `rancher-turtles` chart archives:
++
+[,bash]
+----
+helm pull oci://registry.suse.com/edge/3.1/rancher-turtles-airgap-resources-chart --version 0.3.2
+helm pull oci://registry.suse.com/edge/3.1/rancher-turtles-chart --version 0.3.2
+----
+
+. Transfer the chart archvies to the `management` cluster:
```
e-minguez, line 353: Just me being picky, but as long as you have access to the K8s API, you don't need to do this procedure on the host itself. Both kubectl and helm talk to the API.

asciidoc/day2/migration.adoc
ipetrov117: To be honest I can't really say I agree. Mainly because:

1. We cannot guarantee that the management cluster will be deployed through EIB. Meaning that the Edge helm charts might not even be deployed through the `helm-controller`, heck there might not even be a `helm-controller`. So in this case I would argue that having `HelmChart` resources would be the confusing part.
2. I cannot really justify that being able to do `kubectl get helmcharts` and get all the helm charts (which as described in (1) might not even be the case for non-EIB deployed clusters) can be considered acceptable when we think of the amount of tasks the user would have to go through to deploy the chart..

Regarding (2), what I mean is if the user deployed the resource through a `HelmChart`:

1. First we expect that each `management` 3.0.X `management` cluster would have a `helm-controller` (yes I know EIB requires it, but not all clusters can be EIB created). So the user would have to enable the `helm-controller` or at least validate that it is enabled.
2. The user would the have to create and deploy a `HelmChart` resource.

While I agree that for non-air gapped use-cases (2) might be "okay". I cannot agree that it is okay to expect the user to do the manual steps for an air-gapped environment. Mainly, because the steps involve a very error prone approach.. Namely:

1. Pull the chart (okay)
2. Transfer the chart archive (okay)
3. base64 encode the chart archive (okay)
4. manually copy the chart archive contents to the `HelmChart` file that you will use for deployment (error-prone). - Yes we can automate this, but does it really justify doing this, so that in some use-cases when users do `kubectl get helmcharts` they get all the deployed helm charts? 

 @Kristian-ZH @e-minguez, I just really do not see the value tradeoff in this to be honest. Again if this is a deal breaker and you think that what I said is overcomplicating things, I am happy to change this to a `HelmChart` deployment.

asciidoc/day2/migration.adoc
ipetrov117: You are absolutely correct, I forgot we had the note about `kubectl` being configured to point to the `management` cluster. I will remove this section.

asciidoc/day2/migration.adoc
atanasdinov: I can't help but feel like I also must jump on this discussion.

Using the Helm Controller (and in turn HelmChart resources) was the best or most convenient way at the time to implement airgapped installation of charts via EIB. The way we install those is not really concerning the user and I'd never expect them to run `kubectl get helmcharts` on the cluster.

Using `helm` is much more straightforward and is what people in general are used to in the first place. Trying to "unify" chart upgrades with installations that might not even originate from EIB shouldn't be our goal here.

asciidoc/day2/migration.adoc
ipetrov117: Removed.

asciidoc/day2/migration.adoc
e-minguez: First of all it is not a deal breaker IMHO so whatever is the outcome of this discussion it is ok for me. I understand both points of view, I do.

From an user perspective that have already deployed 3.0 and he is trying to upgrade to 3.1 **the way we said him to do it** (EIB & HelmCharts), moving an internal component to regular `helm install` is confusing. Users are probably aware of the helmcharts objects in their cluster (we don't have any troubleshooting guide but if/when we do, we probably will need to explain this).
My point is that even if we can offer different options we should be consistent and if turtles is deployed from scratch in 3.1 as a helmchart, we should do the same for upgrades as the final picture must be the same.

With regards to this point, shouldn't the user remove the 3.0 manifest for metal3/capi/etc. as well as the clusterctl custom file as part of the upgrade process (I'm talking about this https://github.com/suse-edge/metal3-demo/blob/main/roles/capi/tasks/main.yml) ? or is the rancher-turtles chart smart enough to do it?

atanasdinov: > From an user perspective that have already deployed 3.0 and he is trying to upgrade to 3.1 the way we said him to do it (EIB & HelmCharts), moving an internal component to regular helm install is confusing.

I don't think this is the case. While it is advised to use EIB, nobody is enforcing the user to do so. Also when we're talking about upgrading the platform, we most certainly will not have users perform this operation one or two weeks post initial deployment. They'd most certainly have already deployed and upgraded their workloads multiple times (e.g. using Helm) so I don't think this'd be confusing.

> Users are probably aware of the helmcharts objects in their cluster (we don't have any troubleshooting guide but if/when we do, we probably will need to explain this).

I'm also not convinced about this as I'd expect most users working with the cluster to never have dealt with custom RKE2 (or k3s) resources, but I digress.

ipetrov117: As far as I know `rancher-turtles` handles the upgrade of the `clusterctl` setup. Although I am not an expert on this..

Regarding the consistency point, users can consistently use `helm ls` to get all the charts (even if they are `HelmChart`) resources. Furthermore, there are multiple places where we suggest for core charts to be deployed in the `helm install` manner [(example)](https://suse-edge.github.io/components-kubevirt.html#id-manual-installation-of-edge-virtualization). IMO if we want to do this consistently, we should first discuss it on a wider scope and fix the whole documentation.

Kristian-ZH: I already discussed that with Ivo on DM. Usually, my thinking also includes the future maintenance. 
If the users follow this to upgrade from 3.0 to 3.1 and in the future they want to upgrade to 3.2, then if they followed the suggested approach for setting up cluster (The EIB one), they will have n-1 charts that will be upgraded though changes on the helmcharts (probably using the Upgrade Controller) and just this one that will follow a different approach. This is the thing that will not be good and we will need to bring this case through the entire maintenance window for all releases (3.2,3.3,3.4 and so on...).

But this is not smth so critical, so I agree to proceed in the way that you decide. You can close my comment if you decide to not do this.

ipetrov117: Yea, so even if the user deploys the chart though `helm install`, the `upgrade-controller` will still find the chart and upgrade it. No different approach will have to be taken from 3.1 forwards. 

For 3.0 -> 3.1 user installs `rancher-turtles` through `helm install`
For 3.1 -> 3.X the upgrade will happen through the `upgrade-controller`. Even if the `upgrade-controller` does not support this for 3.2, we still have documentation regarding how to upgrade non-EIB (non-HelmChart) deployed resources.

For 3.1 we will keep the `helm install` approach and from 3.1 onward, the `upgrade-controller` should handle the upgrade, which again will not involve following a different approach, as the controller supports upgrading both `HelmChart` and `helm install` deployed charts.

So, at least for the moment I will close this comment.

Kristian-ZH (approved):

atanasdinov (approved):

Issue #383, opened 2024-09-19
[RFE] SL Micro 6 doesn't allow ssh with password for root user by default (SLE Micro 5.5 does)


e-minguez:
We probably would like to add a small note on the release notes for Edge 3.1 about this.

atanasdinov:
Also look up all `ssh` references in the docs - we might have to adjust EIB definitions to user non-root users.

e-minguez:
SL Micro 6 release notes are going to be updated to reflect those changes as well https://bugzilla.suse.com/show_bug.cgi?id=1230781

PR #464, opened 2024-10-11
3.1 Release Notes update

hardys:
Add feature and bug information, and some upgrade related notes

Co-Authored-By: Rhys Oxenham <rhys.oxenham@suse.com>
Co-Authored-By: Eduardo Mínguez <eduardo.minguez@suse.com>

Fixes: #383

### DIFF
--- a/asciidoc/edge-book/releasenotes.adoc
+++ b/asciidoc/edge-book/releasenotes.adoc
@@ -42,7 +42,7 @@ Summary: SUSE Edge 3.1.0 is the first release in the SUSE Edge 3.1 release strea
 * Management cluster upgrades are now managed via a new Upgrade Controller
 * Edge Image Builder now supports aarch64 (tech preview)
 * Stack Validation results are now published at: https://ci.edge.suse.com
-* nm-configurator updated to utilize nmstate 2.2.38
+* nm-configurator updated to utilize nmstate 2.2.36
 * Edge Image Builder Enhancements:
 ** Adds support for customizing SL Micro 6.0 base images (for SLE Micro 5.5 images, EIB 1.0.x must still be used)
 ** Added the ability to build aarch64 images on an aarch64 host machine (tech preview)
--- a/asciidoc/edge-book/releasenotes.adoc
+++ b/asciidoc/edge-book/releasenotes.adoc
@@ -30,17 +30,57 @@ NOTE: SUSE Edge z-stream releases are tightly integrated and thoroughly tested a
 
 = Release 3.1.0
 
-Availability Date: TBC
+Availability Date: 11th October 2024
 
 Summary: SUSE Edge 3.1.0 is the first release in the SUSE Edge 3.1 release stream.
 
 == New Features
 
-* TBC
+* Updated to SUSE Linux Micro 6.0, Kubernetes 1.30, and Rancher Prime 2.9
+* Updated Cluster API and Metal3/Ironic versions
+* The management cluster CAPI components are now managed via Rancher Turtles
+* Management cluster upgrades are now managed via a new Upgrade Controller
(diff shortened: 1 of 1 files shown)


asciidoc/edge-book/releasenotes.adoc
```diff
@@ -30,17 +30,57 @@ NOTE: SUSE Edge z-stream releases are tightly integrated and thoroughly tested a
...
 == New Features
 
-* TBC
+* Updated to SUSE Linux Micro 6.0, Kubernetes 1.30, and Rancher Prime 2.9
+* Updated Cluster API and Metal3/Ironic versions
+* The management cluster CAPI components are now managed via Rancher Turtles
+* Management cluster upgrades are now managed via a new Upgrade Controller
+* Edge Image Builder now supports aarch64 (tech preview)
+* Stack Validation results are now published at: https://ci.edge.suse.com
+* nm-configurator updated to utilize nmstate 2.2.38
```
atanasdinov, line 45:
```suggestion
* nm-configurator updated to utilize nmstate 2.2.36
```

atanasdinov (approved):
Perhaps we should have a more condensed version of the changes in the future, but I'm fine with this.

hardys:
> Perhaps we should have a more condensed version of the changes in the future, but I'm fine with this.

The initial draft seemed too sparse so after discussion with @rdoxenham we added some more details, open to suggestions if we feel the list is now too verbose, but IMO it's probably OK as it is

asciidoc/edge-book/releasenotes.adoc
```diff
@@ -30,17 +30,57 @@ NOTE: SUSE Edge z-stream releases are tightly integrated and thoroughly tested a
...
-* TBC
+* Updated to SUSE Linux Micro 6.0, Kubernetes 1.30, and Rancher Prime 2.9
+* Updated Cluster API and Metal3/Ironic versions
+* The management cluster CAPI components are now managed via Rancher Turtles
+* Management cluster upgrades are now managed via a new Upgrade Controller
+* Edge Image Builder now supports aarch64 (tech preview)
+* Stack Validation results are now published at: https://ci.edge.suse.com
+* nm-configurator updated to utilize nmstate 2.2.36
+* Edge Image Builder Enhancements:
+** Adds support for customizing SL Micro 6.0 base images (for SLE Micro 5.5 images, EIB 1.0.x must still be used)
```
e-minguez, line 47: I'd say to highlight this somehow. Not sure if moving it to the known issues or not... WDYT?

atanasdinov:
> > Perhaps we should have a more condensed version of the changes in the future, but I'm fine with this.
> 
> The initial draft seemed too sparse so after discussion with @rdoxenham we added some more details, open to suggestions if we feel the list is now too verbose, but IMO it's probably OK as it is

As agreed offline, let's merge and I'll try to come up with a proper list for the EIB ones.

asciidoc/edge-book/releasenotes.adoc
hardys: We discussed this on slack and it's an expected support-matrix constraint, not a known issue - given that we reference 1.1.0 everywhere in this version of the docs it should be OK, but we could mention it in the Upgrade section perhaps if you think it needs clarification?

hardys:
> > > Perhaps we should have a more condensed version of the changes in the future, but I'm fine with this.
> > 
> > 
> > The initial draft seemed too sparse so after discussion with @rdoxenham we added some more details, open to suggestions if we feel the list is now too verbose, but IMO it's probably OK as it is
> 
> As agreed offline, let's merge and I'll try to come up with a proper list for the EIB ones.

Will do - perhaps you could also consider if the comment from @e-minguez should be addressed in that follow-up too?

PR #496, opened 2024-11-20
Cloud hosted mgmt cluster

agracey:


asciidoc/guides/public-cloud.adoc
```diff
@@ -0,0 +1,69 @@
...
+
+TODO: Add diagram here
+
+Downstream clusters need to have access to the Rancher server that's managing it. This means that we need to provide a FQDN that's resolves to a LoadBalancer that routes the traffic to Rancher. 
+
+In the public cloud, this typically is done through the tools provided by your cloud provider.
+
+[NOTE]
+==== 
+For example, in Amazon AWS you could set up an EKS cluster then use an Application Load Balancer (ALB) to both terminate TLS and route traffic to the cluster's ingress controller. If AWS is also your DNS registrar, it can also create a CNAME entry to point at the ALB being created.
```
hardys, line 29: Do we expect to support EKS?  I would expect instead to refer to whatever docs exist for deploying RKE2 on AWS, then configure network connectivity as needed as this is likely to be much closer to the on-premise scenarios we're currently testing/supporting?

asciidoc/guides/public-cloud.adoc
```diff
@@ -0,0 +1,70 @@
...
+- # of workloads 
+- # of extensions being used
+
+More information can be found at https://ranchermanager.docs.rancher.com/v2.10/reference-guides/best-practices/rancher-server/tuning-and-best-practices-for-rancher-at-scale#managing-your-object-counts 
+
+=== Instance Size:
+
+Some basic sizing considerations can be found at https://ranchermanager.docs.rancher.com/v2.10/getting-started/installation-and-upgrade/installation-requirements#rke2-kubernetes
+
+== Federation
```
agracey, line 64: Should this be it's own page? I put it here because it flowed naturally after discussing instance size and I'm not sure if there's enough content to be standalone

cledorze:
Shouldn't we consider adding a link to the list of ports that need to be opened between Rancher in a public cloud and the downstream clusters at the edge? 
Shouldn't we consider adding a link to the list of ports that need to be opened between Rancher in a public cloud and the downstream clusters at the edge?  [Port Requirements – Rancher Manager Docs](https://ranchermanager.docs.rancher.com/getting-started/installation-and-upgrade/installation-requirements/port-requirements)

In any case, it seems to be the right scope. Based on the feedback, we’ll tweak the level of detail if needed.

rdoxenham:
@agracey - is this ready for further review?

hardys:
Closing this as stale

PR #500, opened 2024-11-25
Version attributes - Part 3

jdob:
See #495 for more context.

This PR includes the last of the sweeping changes to use attributes for versions. The bottom of the `versions.adoc` file shows the list of pages that need to be manually reviewed. I included `releasenotes.md` in that list rather than attempt to template it, though I may resurrect the old version matrix page and have it pull values from the variables. 

But this is the last of the major refactoring. Any further work on this should hopefully be less disruptive and not put a pause on others doing docs work.

### DIFF
--- a/asciidoc/components/networking.adoc
+++ b/asciidoc/components/networking.adoc
@@ -79,7 +79,7 @@ We will now ensure that the downloaded base image copy is moved over to the conf
 
 [,shell,subs="attributes"]
 ----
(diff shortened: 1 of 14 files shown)


hardys (approved):
lgtm - I agree we'll need to populate the releasenotes manually since we need to maintain current/latest and also historical versions - perhaps we can consider some kind of automated check in future to ensure the versions and the latest z-stream in the releasenotes align

PR #505, opened 2024-12-16
metal3-quickstart: update known issues

hardys:
The issues mentioned are no longer relevant since 3.1, when we aligned with Rancher Turtles.

Additional information regarding the Turtles integration will be added via a follow-up PR


alknopfler (approved):
LGTM

PR #497, opened 2024-11-20
Version attributes - part 2

jdob:
Second round of updates. See #495 for more information.

PR #506, opened 2024-12-16
Use versioned Rancher docs links

hardys:
Follow-up to #497 where rancher-docs-version was introduced but not applied to all existing links


dprodanov4 (approved):
lgtm!

Issue #471, opened 2024-10-15
[doc] Issue in "Downstream cluster image definition file" - Missing keys and values


peteingithub:
Downstream cluster image definition file:

https://documentation.suse.com/suse-edge/3.1/html/edge/atip-automated-provisioning.html#id-downstream-cluster-image-definition-file-2

`createHomeDir: true`
 
```
packageList:
  - jq
  - openssh-server-config-rootlogin
```

`sccRegistrationCode: "<CODE>" `  # Or other method

Issue #394, opened 2024-09-25
[doc] Issue in "Disabling rebootmgr"


peteingithub:
Disabling rebootmgr:

https://documentation.suse.com/suse-edge/3.0/html/edge/atip-requirements.html#id-disabling-rebootmgr

Actually a better party to disable would be the `transactional-update.timer`. Otherwise the system files would be updated daily, and at some point the system probably reboots to an updated version.

Rebootmgr is also configured here:
https://documentation.suse.com/suse-edge/3.0/html/edge/quickstart-metal3.html#id-downstream-cluster-image-definition-file

PR #509, opened 2024-12-16
metal3: downstream cluster EIB example fixes

hardys:
Ensure jq is installed as it's needed for the manifest rke2-preinstall.service

Also ensure transactional-update timers are disabled since we won't be doing package based updates.

Fixes: #394
Fixes: #471


atanasdinov (approved):

alknopfler (approved):
LGTM

Issue #470, opened 2024-10-15
[doc] Issue in "Downstream cluster image definition file" - Old API version


peteingithub:
Downstream cluster image definition file:

https://documentation.suse.com/suse-edge/3.1/html/edge/atip-automated-provisioning.html#id-downstream-cluster-image-definition-file-2

1.0 -> 1.1

PR #511, opened 2024-12-16
EIB: Align API versions to version-eib-api-latest

hardys:
Although not strictly required it's somewhat confusing having a mix of versions in the docs, so align all references with the version-eib-api-latest attribute for consistency.

Fixes: #470

### DIFF
--- a/asciidoc/components/longhorn.adoc
+++ b/asciidoc/components/longhorn.adoc
@@ -257,7 +257,7 @@ export CONFIG_DIR=$HOME/eib
 mkdir -p $CONFIG_DIR
 
 cat << EOF > $CONFIG_DIR/iso-definition.yaml
-apiVersion: 1.0
+apiVersion: {version-eib-api-latest}
 image:
   imageType: iso
   baseImage: SL-Micro.x86_64-6.0-Base-SelfInstall-GM2.install.iso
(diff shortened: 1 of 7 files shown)


dprodanov4 (approved):

atanasdinov (commented):
I'm hesitating a little about this, as it can very easily run into issues as soon as there's a breaking change in the API definition (i.e. `2.0`). Then we might end up with examples that are not working if we forget to update them. WDYT?

hardys:
> I'm hesitating a little about this, as it can very easily run into issues as soon as there's a breaking change in the API definition (i.e. `2.0`). Then we might end up with examples that are not working if we forget to update them. WDYT?

Yes I had a similar concern, but I guess we'll be pretty cautious to avoid breaking changes unless absolutely needed, and ideally we want the docs to provide a consistent view of a specific API version for each release?

My worry with the current approach is users may not understand the significance of e.g some new example which depends on a newer API version, if we use a mixture of different versions in the docs the risk of breakage/confusion is higher?

When we do make a breaking change we'll have to carefully document it and describe the migration steps, so with this approach we can describe the specific changes for exactly one EIB API version per y-stream Edge release?

atanasdinov (approved):

PR #514, opened 2024-12-18
fix sriov multinode script

alknopfler:
PR Content:

- Fix for the multi-node cluster using the sriov systemd.
- Script reference is being modified in https://github.com/suse-edge/atip/pull/23 so the link will point to the new one


Kristian-ZH (approved):
lgtm

PR #515, opened 2025-01-06
Removing extra quotes in sed command

agracey:
The command as written is broken. I believe this is what's intended?

### DIFF
--- a/asciidoc/integrations/nats.adoc
+++ b/asciidoc/integrations/nats.adoc
@@ -140,7 +140,7 @@ The following command adds `nats` in the build tags to enable the NATS built-in 
 
 [,bash]
 ----
-sed -i '' 's/TAGS="ctrd/TAGS="nats ctrd/g' scripts/build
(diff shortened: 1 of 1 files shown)


hardys:
The fix seems fine, but I'm wondering if we should still document this, and I don't think it's currently fully tested/supported?  @rdoxenham what are your thoughts?

rdoxenham:
Good point @hardys. Let me bring in @bruth - do we want to look at refactoring these docs so they align well with what you are the team are working on? We want to make sure that we're accurately capturing the installation steps but also ensuring that we have good test and support coverage. Thanks!

atanasdinov (commented):
This guide assumes OSX where sed requires the `' '` part.

P.S. I'm fine with updating this but we also have to go back and update the host references and someone would have to retest it.

hardys:
Closing this as stale, I'll follow up so we can decide if this content should be removed in future

PR #523, opened 2025-01-10
IPv6 and dual-stack ATIP documentation updates

mchiappero:
Initial documentation updates (more to follow with multi-node support) regarding IPv6 and dual-stack.

I have added a new section to the release notes to mention the well known issue of the static network configuration not applied in IPA when the DHCPs and RA are also present on the network segment.

PR #533, opened 2025-01-14
Include a proper Tech Preview section in the release notes

mchiappero:
Update the release notes according to recent feedback from https://github.com/suse-edge/suse-edge.github.io/pull/523#issuecomment-2589728503

### DIFF
--- a/asciidoc/edge-book/releasenotes.adoc
+++ b/asciidoc/edge-book/releasenotes.adoc
@@ -130,6 +130,13 @@ registry.suse.com/edge/3.2/kubectl:1.30.3 +
 registry.suse.com/edge/3.2/release-manifest:3.2.0
 |======
 
+= Technology previews
+
+Unless otherwise stated these apply to the 3.2.0 release and all subsequent z-stream versions.
+
+* Akri is a Technology Preview offering, and is not subject to the standard scope of support.
+* Edge Image Builder on aarch64 is a Technology Preview offering, and is not subject to the standard scope of support.
+
 = Components Verification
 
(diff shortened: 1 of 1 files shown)


atanasdinov (approved):
Why did we remove the dual-stack mention, though? It is tech preview for 3.2.

alknopfler (approved):

mchiappero:
> Why did we remove the dual-stack mention, though? It is tech preview for 3.2.

Sorry, I forgot to mention I brought those changes back to #523 (https://github.com/suse-edge/suse-edge.github.io/pull/523/commits/3cb95ec0cc014a7ef2d15fd0f1bae7fad690dbc0).

@fsundermeyer Do we need to add the full implications as you suggested before merging? Either in the same section or as a separate one? Copy and paste? I'd appreciate some guidance here given your expertise :)

fsundermeyer (approved):
LGTM!

fsundermeyer:
> Do we need to add the full implications as you suggested before merging?

Mid term this would be good, IMHO. However. PM and support should first check whether the existing wording also applies to cloud native.

hardys (approved):

hardys:
Lets merge this iteration then we can add more details via follow-up PR's - thanks @mchiappero!

mchiappero:
> > Do we need to add the full implications as you suggested before merging?
> 
> Mid term this would be good, IMHO. However. PM and support should first check whether the existing wording also applies to cloud native.

Maybe we can adopt a trimmed-down version. Will take it offline and ping you back if necessary, for the next releases. Thank you!

PR #534, opened 2025-01-15
Update version-akri-dashboard-extension-chart variable

jtomasek:


jtomasek:
In this case, should we update all charts versions variables to match the versions in release manifest?

hardys:
> In this case, should we update all charts versions variables to match the versions in release manifest?

Yes, thanks for taking care of this! 👍

hardys (approved):

atanasdinov (approved):

PR #536, opened 2025-01-15
[3.2] Edge 3.2 Day2 migration documentation

ipetrov117:
Introduces a simplified Edge `3.2` migration doc.

Notes:
- version attributes have been added to the `adoc` itself as opposed to adding them to `versions.adoc` because these attributes will only be specific to the migration doc.
- Metal3 version has purposely not been templated - it needs to be always relevant to the Edge `3.2.0` release. The whole section will probably not be needed for Edge `3.3` migrations.


asciidoc/day2/migration.adoc
```diff
@@ -11,1005 +11,159 @@ ifdef::env-github[]
...
+The only difference is that your `UpgradePlan` *must* specify the `{static-edge-version}` release version:
 
-[,bash]
+[,bash,subs="attributes"]
 ----
-helm install rancher-turtles oci://registry.suse.com/edge/3.1/rancher-turtles-chart --version 0.3.2 --namespace rancher-turtles-system --create-namespace
+apiVersion: lifecycle.suse.com/v1alpha1
+kind: UpgradePlan
+metadata:
+  name: upgrade-plan-mgmt-{dash-separated-static-edge-version}
```
atanasdinov, line 116: This goes against our assumption to stick with a single UpgradePlan resource in a cluster. Is it intended?

asciidoc/day2/migration.adoc
```diff
@@ -11,1005 +11,159 @@ ifdef::env-github[]
...
-To migrate the CAPI controllers versions to `Edge 3.1.0` compatible versions, install the `Rancher Turtles` chart:
+The only difference is that your `UpgradePlan` *must* specify the `{static-edge-version}` release version:
 
-[,bash]
+[,bash,subs="attributes"]
 ----
-helm install rancher-turtles oci://registry.suse.com/edge/3.1/rancher-turtles-chart --version 0.3.2 --namespace rancher-turtles-system --create-namespace
+apiVersion: lifecycle.suse.com/v1alpha1
+kind: UpgradePlan
+metadata:
```
atanasdinov, line 115: It's missing the namespace.

asciidoc/day2/migration.adoc
```diff
@@ -11,1005 +11,159 @@ ifdef::env-github[]
...
 
-After some time, the controller pods running in the `capi-system`, `capm3-system`, `rke2-bootstrap-system` and `rke2-control-plane-system` namespaces are upgraded with the `Edge 3.1.0` compatible controller versions.
-
-For information on how to install `Rancher Turtles` in an air-gapped environment, refer to <<day2-migration-mgmt-helm-capi-air-gapped, Rancher Turtles air-gapped installation>>.
-
-[#day2-migration-mgmt-helm-capi-air-gapped]
-===== Rancher Turtles air-gapped installation
-
-[NOTE]
+[IMPORTANT]
```
atanasdinov, line 121: Is this section required? The last sentence already mentions that a `3.2.0` value must be specified.

asciidoc/day2/migration.adoc
```diff
@@ -11,1005 +11,159 @@ ifdef::env-github[]
...
-
-[#day2-migration-mgmt-helm-eib]
-==== Edge Helm chart upgrade - EIB
-
-This section explains how to upgrade a Helm chart from the Edge component stack, deployed via <<components-eib, EIB>>, to an `Edge 3.1.0` compatible version.
+For information on how to use the above `UpgradePlan` to do a migration, refer to <<{cluster-type}-day2-upgrade-controller, Upgrade Controller upgrade process>>.
 
-===== Prerequisites
+[#day2-migration-mgmt-fleet]
+=== Fleet
```
atanasdinov, line 129: Should we leave a note that this is only for airgapped upgrades?

asciidoc/day2/migration.adoc
ipetrov117: You are right, I'll remove this suffix from here and from the `Upgrade Controller` documentation as well.

asciidoc/day2/migration.adoc
ipetrov117: I'll add it, however it will not be a "hardcoded" value. It will be more like:

```yaml
..
namespace: <upgrade_controller_namespace>
..
```

As we do not have a robust way to ensure that `Upgrade Controller` is always installed in the same namespace.

asciidoc/day2/migration.adoc
ipetrov117: I wanted to stress this further, but I see how it can be seen as repetitive and not really add value. Will remove it.

asciidoc/day2/migration.adoc
ipetrov117: Opted for the `CHANGE_ME` approach as having `<>` in a block that has `subs="attributes"` causes `daps2docker` to fail.

atanasdinov (dismissed):

asciidoc/day2/migration.adoc
```diff
@@ -11,1005 +11,161 @@ ifdef::env-github[]
...
+helm pull oci://registry.suse.com/edge/{version-edge-registry}/metal3-chart --version 302.0.0+up0.9.0
 ----
 
-. Mark the node as schedulable:
+. Manually apply the BMO CRDs:
 +
 [,bash]
 ----
-kubectl uncordon <node_name>
+tar -xvf metal3-chart-302.0.0+up0.9.0.tgz && kubectl apply -f metal3/charts/baremetal-operator/crds/
```
hardys, line 68: Thanks for taking care of this!

Looks fine, but if we wanted a more compact version, we could do e.g:

```
helm show crds oci://registry.suse.com/edge/3.2/metal3-chart --version 302.0.0+up0.9.0 | kubectl apply -f -
```

hardys (approved):

asciidoc/day2/migration.adoc
ipetrov117: Agreed, I'll take care of this ASAP.

hardys (approved):

atanasdinov (approved):

PR #544, opened 2025-01-17
Rename ATIP to SUSE Edge for Telco

hardys:
To align with recent branding changes update the ATIP naming

To reduce the risk of breakage this doesn't update any filenames or links, we'll have to adjust them via follow-up changes.

### DIFF
--- a/asciidoc/product/atip-requirements.adoc
+++ b/asciidoc/product/atip-requirements.adoc
@@ -13,14 +13,14 @@ endif::[]
 
 === Hardware
 
-The hardware requirements for the ATIP nodes are based on the following components:
+The hardware requirements SUSE Edge for telco are as follows:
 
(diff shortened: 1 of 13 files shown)


asciidoc/product/atip-requirements.adoc
```diff
@@ -13,14 +13,14 @@ endif::[]
 
 === Hardware
 
-The hardware requirements for the ATIP nodes are based on the following components:
+The hardware requirements SUSE Edge for telco are as follows:
```
atanasdinov, line 16:
```suggestion
The hardware requirements SUSE Edge for Telco are as follows:
```

atanasdinov (approved):

mchiappero (approved):
One thing left to do is to rename the files and update the references.

atanasdinov (approved):

hardys:
> One thing left to do is to rename the files and update the references.

As mentioned in the description I'm going to leave that as a follow-up, as I think the risk of breaking things is too much this close to release, also if we modify the anchor naming it will break any external links to the docs compared to previous versions (perhaps this is unavoidable but I think we need more time to consider how to handle it)

Issue #528, opened 2025-01-13
Add a reference to the toolbox 5.5 vs 6.0 issue on the release notes


e-minguez:
https://github.com/SUSE/release-notes/issues/6

Basically SL Micro 6.0 toolbox need to use 5.5 toolbox instead. We probably would like to add a note on the known issues of the release notes.

PR #546, opened 2025-01-20
release-notes: Add note regarding toolbox image

hardys:
Make a note that the suse/sle-micro/5.5/toolbox image may be required

In future we can add more comprehensive troubleshooting docs, but this at least points in the right direction for any users who encounter issues trying to use the 6.0 toolbox image

Fixes: #528

### DIFF
--- a/asciidoc/edge-book/releasenotes.adoc
+++ b/asciidoc/edge-book/releasenotes.adoc
@@ -49,6 +49,7 @@ Summary: SUSE Edge 3.2.0 is the first release in the SUSE Edge 3.2 release strea
 == Known issues
 
 * When deploying via the directed network provisioning flow, a bug affects clusters with static IPs in networks with DHCP servers and/or RAs: static network configurations only apply to the provisioned host and will not be in effect during the host discovery and enrollment. Please refer to the https://github.com/suse-edge/atip/tree/main/telco-examples/edge-clusters/dhcp-less/dual-stack/single-node#readme[SUSE Edge for Telco examples repository] for more details and updates.
+* When using `toolbox` in SUSE Linux Micro 6.0, the default container image does not contain some tools which were included in the previous 5.5 version.  The workaround is to configure toolbox to use the previous `suse/sle-micro/5.5/toolbox` container image, see `toolbox --help` for options to configure the image.
 
 == Components Versions
 


asciidoc/edge-book/releasenotes.adoc
```diff
@@ -49,6 +49,7 @@ Summary: SUSE Edge 3.2.0 is the first release in the SUSE Edge 3.2 release strea
 == Known issues
 
 * When deploying via the directed network provisioning flow, a bug affects clusters with static IPs in networks with DHCP servers and/or RAs: static network configurations only apply to the provisioned host and will not be in effect during the host discovery and enrollment. Please refer to the https://github.com/suse-edge/atip/tree/main/telco-examples/edge-clusters/dhcp-less/dual-stack/single-node#readme[SUSE Edge for Telco examples repository] for more details and updates.
+* When using `toolbox` in SUSE Linux Micro 6.0, the default container image does not contain some tools which were included in the previous 5.5 version.  The workaround is to configure toolbox to use the previous `suse/sle-micro/5.5/toolbox` container image.
```
e-minguez, line 52:
```suggestion
* When using `toolbox` in SUSE Linux Micro 6.0, the default container image does not contain some tools which were included in the previous 5.5 version.  The workaround is to configure toolbox to use the previous `suse/sle-micro/5.5/toolbox` container image:
[,console]
----
$ cat << EOF > ~/.toolboxrc
IMAGE=suse/sle-micro/5.5/toolbox:latest
EOF
----

e-minguez: Does it make sense to add this?

e-minguez (dismissed):
LGTM, just added a suggestion in case you want to accept it.

asciidoc/edge-book/releasenotes.adoc
hardys: I could just say "see `toolbox --help` for options to configure the image`  as this mentions e.g `--image` and also the configuration file approach?

hardys: I added a note to clarify this, but referenced `--help` instead of describing the configuration file approach

diconico07 (approved):

e-minguez (approved):

Issue #557, opened 2025-01-24
[doc] Issue in "Downstream cluster provisioning with Directed network provisioning (multi-node)" -  RolloutStrategy missing from RKE2ControlPlane manifest


peteingithub:
```
  rolloutStrategy:
    type: "RollingUpdate"
    rollingUpdate:
      maxSurge: 0
```

PR #504, opened 2024-12-11
change v1alpha1 from capi controlplane and minor changes

alknopfler:
- Change to move from v1alpha1 to v1beta1 in rke2controlplane 
- version in spec out of the agentConfig
2nd EDIT:
- Add the identifier mac to the bmh spec file
- remove OS version from the images to make it easy to be updated

PR #559, opened 2025-01-27
Add missing RKE2ControlPlane rolloutStrategy

hardys:
This is required since moving to the v1beta1 API in #504

In future we should add more details describing the options but for now this ensures the examples are functional, and aligned with the ATIP repo examples.

Fixes: #557


alknopfler (dismissed):
LGTM. Good catch! Thanks

atanasdinov (approved):

atanasdinov (approved):

Issue #551, opened 2025-01-23
[doc] Issue in "Downstream cluster provisioning with Directed network provisioning (multi-node)" - Typo


peteingithub:
multinode-node-cluster-controlplane-template -> multinode-cluster-controlplane-template

Issue #550, opened 2025-01-23
[doc] Issue in "Downstream cluster provisioning with Directed network provisioning (multi-node)" - Cluster name mismatch


peteingithub:
clusterName: single-node-cluster -> clusterName: multinode-cluster

PR #567, opened 2025-01-28
Downstream cluster provisioning fix typos

hardys:
multinode-node-cluster-controlplane-template -> multinode-cluster-controlplane-template
clusterName: single-node-cluster -> clusterName: multinode-cluster

Fixes: #551
Fixes: #550

### DIFF
--- a/asciidoc/product/atip-automated-provision.adoc
+++ b/asciidoc/product/atip-automated-provision.adoc
@@ -1231,10 +1231,10 @@ The `Metal3DataTemplate` object specifies the `metaData` for the downstream clus
 apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
 kind: Metal3DataTemplate
 metadata:
-  name: multinode-node-cluster-controlplane-template
+  name: multinode-cluster-controlplane-template
   namespace: default
 spec:
-  clusterName: single-node-cluster
+  clusterName: multinode-cluster
   metaData:
     objectNames:
       - key: name


asciidoc/product/atip-automated-provision.adoc
```diff
@@ -1231,10 +1231,10 @@ The `Metal3DataTemplate` object specifies the `metaData` for the downstream clus
 apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
 kind: Metal3DataTemplate
 metadata:
-  name: multinode-node-cluster-controlplane-template
+  name: multinode-cluster-controlplane-template
   namespace: default
 spec:
-  clusterName: single-node-cluster
+  clusterName: multinode-cluster
```
atanasdinov, line 1237: Is this the correct name? Should we adjust the references to `multi-note-...` instead. Applies for the `metadata.name` too.

atanasdinov (approved):

asciidoc/product/atip-automated-provision.adoc
hardys: This refers to the `Cluster` resource so I think it's correct?

```
apiVersion: cluster.x-k8s.io/v1beta1
kind: Cluster
metadata:
  name: multinode-cluster
  namespace: default
```

AFAICS the resource names are consistently using `multinode` (except this one), the text does use `multi-node` which arguably reads a little better so I wasn't planning to update that (or at least not in this PR)

Issue #548, opened 2025-01-23
[doc] Issue in "Known issues" - Endpoint Copier version mismatch?


peteingithub:
Should the version be 0.2.1 instead of 0.2.0?

hardys:
As discussed in #568 this version of the chart (0.2.1) [does not match the app version](https://github.com/suse-edge/charts/blob/main/charts/endpoint-copier-operator/0.2.1/Chart.yaml) (0.2.0) so the docs are correct

PR #568, opened 2025-01-28
Fix endpoint-copier-operator versions

hardys:
Fixes: #548


hardys:
> This is not correct. ECO is at [version 0.2.0](https://github.com/suse-edge/endpoint-copier-operator/releases/tag/v0.2.0) with chart version 0.2.1.

Ah I missed that the chart and app version don't align in [this version of the chart](https://github.com/suse-edge/charts/blob/main/charts/endpoint-copier-operator/0.2.1/Chart.yaml)

PR #577, opened 2025-02-04
UpgradeController and SystemUpgradeController

jiwonhu:
I am hoping this PR clarifies the purpose of two-similarly-named-controllers and answer the basic questions from the users, such as "why both controllers are needed" and their individual roles.

* Adds "UC vs SUC" to explain when to use which controller
* Explains the use of SUC and Fleet (instead of UC) for managing downstream clusters, despite UC's original development for SUSE Edge
* Specifies the minimum required components for the upgrade. This considers that many customers don't install the full suite of SUSE Edge

### DIFF
--- a/asciidoc/components/system-upgrade-controller.adoc
+++ b/asciidoc/components/system-upgrade-controller.adoc
@@ -19,7 +19,11 @@ ____
 
 == How does SUSE Edge use System Upgrade Controller?
 
-*SUC* is used to assist in the various "Day 2" operations that need to be executed in order to upgrade management/downstream clusters from one Edge platform version to another. Day 2 operations are defined in the form of *SUC Plans*. Based on the these plans, SUC deploys workloads on each node that executes the respective Day 2 operations.
+The SUC facilitates various "day 2" operations, including SUSE Edge version upgrades. SUSE Edge uses it to identifie and upgrade all existing components (OS, Kubernetes, etc.) within a cluster.
+
+While SUC is used to upgrade downstream clusters directly, it also operates behind the scenes when the Upgrade Controller updates management clusters.
+
+Day 2 operations are defined in the form of *SUC Plans*. Based on the these plans, SUC deploys workloads on each node that executes the respective Day 2 operations.
 
 [#components-system-upgrade-controller-install]
 == Installing the System Upgrade Controller
--- a/asciidoc/components/upgrade-controller.adoc
+++ b/asciidoc/components/upgrade-controller.adoc
@@ -21,6 +21,15 @@ A Kubernetes controller capable of performing infrastructure platform upgrades c
 * Additional components (Rancher, Elemental, SUSE Security, etc.)
 ____
 
+The Upgrade Controller intelligently upgrades a SUSE Edge *management cluster* to a new version. It identifies and upgrades all existing components, including the OS, Kubernetes, and others. If a component's version remains unchanged in the new SUSE Edge release, it will not be upgraded. Also, any components not found in the cluster are skipped during the upgrade process. It uses System Upgrade Controller (SUC) under the hood.
+
+
+== Upgrade Controller vs System Upgrade Controller
+
+While the System Upgrade Controller (SUC) is a general-purpose tool for Kubernetes node upgrades, it isn't tailored for SUSE Edge. Therefore, the Upgrade Controller (UC) was created to provide a streamlined upgrade experience specifically for SUSE Edge clusters.  UC manages upgrades between SUSE Edge versions (e.g., 3.1.2 to 3.2.0 or 3.1.1 to 3.2.2), simplifying management by treating the entire platform as a single unit.
+
+However, the current design of Cluster API prevents Upgrade Controller functioning correctly for downstream clusters. Cluster API separates OS provisioning from Kubernetes deployment and requires complete node replacement. Therefore, in current release, SUSE Edge utilizes SUC and Fleet for downstream cluster upgrades. Management clusters can be upgraded with the UC.
+
 == How does SUSE Edge use Upgrade Controller?
 
 The *Upgrade Controller* is essential in automating the (formerly manual) "Day 2" operations required to upgrade management clusters from one SUSE Edge release version to the next.
--- a/asciidoc/day2/migration.adoc
+++ b/asciidoc/day2/migration.adoc
@@ -24,6 +24,24 @@ Always perform cluster migrations from the `latest Z-stream` release of `Edge {p
 Always migrate to the `Edge {static-edge-version}` release. For subsequent post-migration upgrades, refer to the <<day2-mgmt-cluster, management>> and <<day2-downstream-clusters, downstream>> cluster sections.
 ====
 
+== Upgrade Prerequisites
+We understand you may not use all SUSE Edge components. The following components are required for a successful upgrade:
+
+* SUSE Linux Micro
+* k3s or RKE2 as Kubernetes platform
+* If downstream cluster is upgraded: System Upgrade Controller on each downstream cluster. Fleet on management cluster is necessary to trigger the upgrade
+* If management cluster is upgraded:
+** If cluster is not air-gapped, Upgrade Controller
+** If cluster is air-gapped, Fleet
+
+
+[NOTE]
+====
+Question. Are both UC and SUC necessary?
+
+Due to the current Cluster API design, the Upgrade Controller (UC) cannot upgrade downstream clusters. Cluster API's separation of OS provisioning from Kubernetes deployment, coupled with its requirement for complete node replacement, prevents UC from functioning correctly in this context. Therefore, this release utilizes SUC and Fleet for downstream cluster upgrades. Management clusters can be upgraded with the UC.
+====
+
 [#day2-migration-mgmt]
 == Management Cluster
 :cluster-type: management


asciidoc/day2/migration.adoc
```diff
@@ -24,6 +24,24 @@ Always perform cluster migrations from the `latest Z-stream` release of `Edge {p
 Always migrate to the `Edge {static-edge-version}` release. For subsequent post-migration upgrades, refer to the <<day2-mgmt-cluster, management>> and <<day2-downstream-clusters, downstream>> cluster sections.
 ====
 
+== Upgrade Prerequisites
+We understand you may not use all SUSE Edge components. The following components are required for a successful upgrade:
+
+* SUSE Linux Micro
+* k3s or RKE2 as Kubernetes platform
+* If downstream cluster is upgraded: System Upgrade Controller on each downstream cluster. Fleet on management cluster is necessary to trigger the upgrade
```
ipetrov117, line 32: Most of this is covered in the "Day 2" section for both [Management](https://suse-edge.github.io/day2-mgmt-cluster.html#management-day2-fleet-components) and [Downstream](https://suse-edge.github.io/day2-downstream-clusters.html#downstream-day2-fleet-components) cluster upgrades. The only thing that is missing is a "Fleet" component in the "Management" cluster section, it was omitted with the thinking that users would be aware that Fleet needs to be only on the Management cluster.  

If you feel like this information needs to be added, I suggest adding it under [here](https://github.com/suse-edge/suse-edge.github.io/blob/main/asciidoc/day2/fleet.adoc?plain=1#L26).

asciidoc/day2/migration.adoc
```diff
@@ -24,6 +24,24 @@ Always perform cluster migrations from the `latest Z-stream` release of `Edge {p
...
+* If management cluster is upgraded:
```
ipetrov117, line 33: Isn't this [note](https://github.com/suse-edge/suse-edge.github.io/blob/main/asciidoc/day2/migration.adoc?plain=1#L69) enough to instruct users that they shouldn't use the Upgrade Controller for air-gapped environments?

If it is not, I would suggest to enhance it instead of having this part here.

asciidoc/day2/migration.adoc
```diff
@@ -24,6 +24,24 @@ Always perform cluster migrations from the `latest Z-stream` release of `Edge {p
...
+** If cluster is not air-gapped, Upgrade Controller
+** If cluster is air-gapped, Fleet
+
+
+[NOTE]
```
ipetrov117, line 38: I feel like this is a misleading question. To me it suggest we use both Upgrade Controller and SUC for a single upgrade/migration approach, which is not the case. SUC is used in the Fleet use-case. Upgrade Controller is its own use-case altogether.

If users have this question it will be later down the line and they can easily find an answer to it but looking at the Upgrade Controller or SUC component docs that are referenced in all "Day 2" sections.

IMO we should not try to answer this question in every section, it should be enough to only have an answer in the Upgrade Controller and SUC component docs. Let me know what you think.

asciidoc/day2/migration.adoc
```diff
@@ -24,6 +24,24 @@ Always perform cluster migrations from the `latest Z-stream` release of `Edge {p
 Always migrate to the `Edge {static-edge-version}` release. For subsequent post-migration upgrades, refer to the <<day2-mgmt-cluster, management>> and <<day2-downstream-clusters, downstream>> cluster sections.
 ====
 
+== Upgrade Prerequisites
+We understand you may not use all SUSE Edge components. The following components are required for a successful upgrade:
+
+* SUSE Linux Micro
```
ipetrov117, line 30: I do not feel like we need to explicitly state this here. To me OS and K8s components are requirements for an initial SUSE Edge platform setup. IMO listing them here is telling the user something that they should already know.

asciidoc/day2/migration.adoc
```diff
@@ -24,6 +24,24 @@ Always perform cluster migrations from the `latest Z-stream` release of `Edge {p
 Always migrate to the `Edge {static-edge-version}` release. For subsequent post-migration upgrades, refer to the <<day2-mgmt-cluster, management>> and <<day2-downstream-clusters, downstream>> cluster sections.
 ====
 
+== Upgrade Prerequisites
```
ipetrov117, line 27: Added some comments below, but generally I am not sure we need this section. Let me know what you think.

ipetrov117 (commented):
Hey @jiwonhu, thanks for doing this. Will be happy to approve once we discuss the comments below 🚀

asciidoc/day2/migration.adoc
jiwonhu: Edge SKU gives the access to so many products/components and this hugely confuses the customers and field teams. I've gotten feedback that they are unclear about which components are mandatory versus optional, and whether users are expected to utilize all included products. 
Given this is a "day 2" feature and the end users have different install bases, I believe we need to clearly define the minimal required components. Without this, the users cannot tell whether they can perform the upgrade with their install base or what's required to perform the upgrade.

asciidoc/day2/migration.adoc
jiwonhu: Responded in above comment

asciidoc/components/upgrade-controller.adoc
```diff
@@ -21,6 +21,15 @@ A Kubernetes controller capable of performing infrastructure platform upgrades c
 * Additional components (Rancher, Elemental, SUSE Security, etc.)
 ____
 
+The Upgrade Controller intelligently upgrades a SUSE Edge *management cluster* to a new version. It identifies and upgrades all existing components, including the OS, Kubernetes, and others. If a component's version remains unchanged in the new SUSE Edge release, it will not be upgraded. Also, any components not found in the cluster are skipped during the upgrade process. It uses System Upgrade Controller (SUC) under the hood.
+
+
+== Upgrade Controller vs System Upgrade Controller
```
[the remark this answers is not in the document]
jiwonhu: I agree that "how does SUSE Edge use Upgrade Controller" better be at the entry point.

asciidoc/components/upgrade-controller.adoc
```diff
@@ -21,6 +21,15 @@ A Kubernetes controller capable of performing infrastructure platform upgrades c
...
+
+While the System Upgrade Controller (SUC) is a general-purpose tool for Kubernetes node upgrades, it isn't tailored for SUSE Edge. Therefore, the Upgrade Controller (UC) was created to provide a streamlined upgrade experience specifically for SUSE Edge clusters.  UC manages upgrades between SUSE Edge versions (e.g., 3.1.2 to 3.2.0 or 3.1.1 to 3.2.2), simplifying management by treating the entire platform as a single unit.
```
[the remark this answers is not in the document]
jiwonhu: I understand the naming is unfortunate but I like the name 'Upgrade Controller' :-) I will avoid abbreviating it to UC.
This section was trickiest for me to write. I accept the suggestion.

asciidoc/day2/migration.adoc
jiwonhu: The idea of this section is giving a short list of prerequisites in the beginning of the section so the end user can check. I believe it is a common practice, [this is one example](https://learn.microsoft.com/en-us/mem/configmgr/sum/plan-design/prerequisites-for-software-updates). We can't expect user to figure all the prereqs out by going through the upgrade steps.

asciidoc/components/upgrade-controller.adoc
```diff
@@ -21,6 +21,15 @@ A Kubernetes controller capable of performing infrastructure platform upgrades c
 * Additional components (Rancher, Elemental, SUSE Security, etc.)
 ____
 
+The Upgrade Controller intelligently upgrades a SUSE Edge *management cluster* to a new version. It identifies and upgrades all existing components, including the OS, Kubernetes, and others. If a component's version remains unchanged in the new SUSE Edge release, it will not be upgraded. Also, any components not found in the cluster are skipped during the upgrade process. It uses System Upgrade Controller (SUC) under the hood.
```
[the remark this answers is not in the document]
jiwonhu: The intention of this paragraph is to give a short summary of Upgrade Controller. I went through this page and could not fully understand. I think one of the reasons is our documentation gives out the links, instead we could give a brief information then give the link.

I agree the first sentence (it identifies...) and last sentence (It uses SUC under the hood) can be gone. But "components versions unchanged" is what end user would want to know and I don't want to hide it in the back.

asciidoc/components/upgrade-controller.adoc
```diff
@@ -21,6 +21,15 @@ A Kubernetes controller capable of performing infrastructure platform upgrades c
...
+
+
+== Upgrade Controller vs System Upgrade Controller
+
+While the System Upgrade Controller (SUC) is a general-purpose tool for Kubernetes node upgrades, it isn't tailored for SUSE Edge. Therefore, the Upgrade Controller (UC) was created to provide a streamlined upgrade experience specifically for SUSE Edge clusters.  UC manages upgrades between SUSE Edge versions (e.g., 3.1.2 to 3.2.0 or 3.1.1 to 3.2.2), simplifying management by treating the entire platform as a single unit.
+
+However, the current design of Cluster API prevents Upgrade Controller functioning correctly for downstream clusters. Cluster API separates OS provisioning from Kubernetes deployment and requires complete node replacement. Therefore, in current release, SUSE Edge utilizes SUC and Fleet for downstream cluster upgrades. Management clusters can be upgraded with the UC.
```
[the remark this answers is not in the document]
jiwonhu: I accept that the sentence could be understood that way, although it was not my intention. Sorry!

What we provide is quite complicated. so I wanted to clarify: "There was SUC, but it wasn't enough for edge, so we introduced UpgradeController. But why can't we use this awesome tool for downstream clusters, where it is most needed? Because of current implementation of CAPI."

End users need to know this upfront, instead of discovering from the bottom of the doc. To clarify why we introduced two features of similar name. Well, that's my view. Otherwise we could simply state: "Use UpgController for mgmt, use SUC for downstream, don't ask why."

I am good to remove the 1st paragraph (yours sounds a lot better) but don't want to give up on explaining why we chose SUC and Fleet for downstream clusters, despite the introduction of UpgController.

asciidoc/day2/migration.adoc
jiwonhu: I accept this comment and can remove this NOTE, if we can keep the "UC vs SUC" section in upgrade controller section.

asciidoc/day2/migration.adoc
jiwonhu: Hi @ipetrov117 and @atanasdinov, let me see if I can explain this more clearly. SUSE Edge bundles numerous components, but user deployments vary significantly. Some might use a minimal setup (e.g. k3s + Rancher) without SLEM due to their preference of other OS (SLEM and immutable OS are not that popular). While others may use more components (e.g. SLEM+k3s+Rancher without utilizing Fleet).
I find current doc doesn't explicitly list the minimum required components for an upgrade. This makes impossible for the users to determine if their specific setup is eligible for an upgrade. For example, a user running products X, Y, and Z has no way to know, based on the current doc, whether they can upgrade their downstream/management cluster. Since many components are optional, the upgrade doc must clarify how these optional components affect the upgrade process and what the min requirements are.

Please let me know if you still think this section is unnecessary.

asciidoc/day2/migration.adoc
jiwonhu: Removed the [NOTE] section with a new commit.

jiwonhu:
Updated based on the feedback received. I still feel `Upgrade Prerequisites` is necessary, I responded in above thread, let me see if the team still feels otherwise.
I squashed 4 commits, hope that is okay? I am still learning git...

asciidoc/day2/migration.adoc
hardys: > Some might use a minimal setup (e.g. k3s + Rancher) without SLEM due to their preference of other OS

jiwonhu: Understood. What's the team's verdict on this section? Am I alone thinking it is important to list the minimum required components for an upgrade, when we offer 20+ components and not indicate which one is optional?

atanasdinov: Well, if we stick with the idea that SUSE Linux Micro and either k3s or RKE2 are always there, everything else is documented.

Installing the Upgrade Controller already has its own [prerequisites](https://documentation.suse.com/suse-edge/3.2/html/edge/components-upgrade-controller.html#components-upgrade-controller-installation). The migration (3.1 -> 3.2) documentation also already has [prerequisites](https://documentation.suse.com/suse-edge/3.2/html/edge/day2-migration.html#day2-migration-mgmt-upgrade-controller-prereq).

The Upgrade Controller is fully capable of understanding what's on the cluster and what isn't i.e. if any of the other components (e.g. Longhorn) is not on the cluster, it will be skipped.

asciidoc/day2/migration.adoc
jiwonhu: Withdrew this section.

asciidoc/day2/migration.adoc
jiwonhu: Withdrew this section.

asciidoc/day2/migration.adoc
jiwonhu: Withdrew this section as per team's feedback.

asciidoc/day2/migration.adoc
jiwonhu: Withdrew this section as per team's feedback.

asciidoc/components/upgrade-controller.adoc
```diff
@@ -25,12 +25,25 @@ ____
...
+[#components-upgrade-controller-uc-vs-suc]
+== Upgrade Controller vs System Upgrade Controller
+
+The <<components-system-upgrade-controller, System Upgrade Controller (SUC)>> is a general-purpose tool that propagates upgrade instructions to specific Kubernetes nodes.
+
+While it supports some "Day 2" operations for the SUSE Edge platform, it *does not* cover all of them. Moreover, even for supported operations, users have to manually configure, maintain, and deploy multiple `SUC Plans` — an error-prone process that can lead to unexpected issues.
+
+This led to the need for a tool that **automates** and **abstracts** the complexity of managing various "Day 2" operations for the SUSE Edge platform. Thus, the `Upgrade Controller` was developed. It simplifies the upgrade process by introducing a single `user-facing resource` that drives the upgrade. Users only need to manage this resource, while the `Upgrade Controller` takes care of the rest.
+
+Management clusters are upgraded via the `Upgrade Controller`; however, downstream clusters are upgraded by SUC and Fleet in current release.
```
ipetrov117, line 44: To be honest I am not sure we need this here. Isn't it enough to have a note in the Upgrade Controller description, saying it currently works only for non air-gapped clusters (see my previous comment)?

In this section, I would prefer to avoid diving into the whole complexity around what tools we support for upgrade in which use-cases, as this is described in several other places (e.g. the Upgrade Controller description, the [migration documentation](https://suse-edge.github.io/day2-migration.html#day2-migration-mgmt-upgrade-controller), the [upgrade documentation](https://suse-edge.github.io/day2-mgmt-cluster.html#management-day2-upgrade-controller)).  WDYT?

ipetrov117 (commented):
Hey @jiwonhu, the documentation is looking good. I left out some comments, will be happy to hear your thoughts. Thank you!

asciidoc/components/upgrade-controller.adoc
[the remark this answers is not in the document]
ipetrov117: Hey @jiwonhu, I totally agree that the Upgrade Controller description needs to give a better overview of the tool. Check [this](https://github.com/suse-edge/suse-edge.github.io/pull/577#discussion_r1952559109) comment about a suggestion of how the description could look like and let me know what you think.

asciidoc/components/upgrade-controller.adoc
```diff
@@ -25,12 +25,25 @@ ____
 
 The *Upgrade Controller* is essential in automating the (formerly manual) "Day 2" operations required to upgrade management clusters from one SUSE Edge release version to the next.
 
-To achieve this automation, the Upgrade Controller utilizes tools such as the <<components-system-upgrade-controller, System Upgrade Controller>> and the link:https://github.com/k3s-io/helm-controller/[Helm Controller]. 
+To achieve this automation, the Upgrade Controller utilizes tools such as the <<components-system-upgrade-controller, System Upgrade Controller>> and the link:https://github.com/k3s-io/helm-controller/[Helm Controller].
 
 For further details on how the Upgrade Controller works, see <<components-upgrade-controller-how>>.
 
 For known limitations that the Upgrade Controller has, see <<components-upgrade-controller-known-issues>>.
```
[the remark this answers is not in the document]
jiwonhu: Sounds good, added.

asciidoc/components/system-upgrade-controller.adoc
```diff
@@ -19,7 +19,15 @@ ____
...
 == How does SUSE Edge use System Upgrade Controller?
 
-*SUC* is used to assist in the various "Day 2" operations that need to be executed in order to upgrade management/downstream clusters from one Edge platform version to another. Day 2 operations are defined in the form of *SUC Plans*. Based on the these plans, SUC deploys workloads on each node that executes the respective Day 2 operations.
+SUSE Edge uses `SUC` to facilitate various "Day 2" operations related to OS and Kubernetes version upgrades on management and downstream clusters.
+
+"Day 2" operations are defined through `SUC Plans`. Based on these plans, `SUC` deploys workloads on each node to execute the respective "Day 2" operation.
+
+`SUC` is also used within the <<components-upgrade-controller>>. To learn more about the key differences between SUC and the Upgrade Controller, see <<components-upgrade-controller-uc-vs-suc>>.
+
+While SUC is used to upgrade downstream clusters directly, it also operates behind the scenes when the Upgrade Controller updates management clusters.
```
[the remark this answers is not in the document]
jiwonhu: Removed, thanks!

asciidoc/components/system-upgrade-controller.adoc
```diff
@@ -19,7 +19,15 @@ ____
...
+
+Day 2 operations are defined in the form of *SUC Plans*. Based on the these plans, SUC deploys workloads on each node that executes the respective Day 2 operations.
```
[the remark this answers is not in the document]
jiwonhu: Removed line 30 as discussed, thanks!

asciidoc/components/upgrade-controller.adoc
jiwonhu: Agree that what's mentioned in line 44 is mentioned in other pages. Removed as per our discussion.

ipetrov117 (dismissed):
Hey @jiwonhu, awesome job! I left some minor "nits" regarding the additional newlines that were added, but other than that, it looks good to me!

asciidoc/components/system-upgrade-controller.adoc
```diff
@@ -19,7 +19,12 @@ ____
 
 == How does SUSE Edge use System Upgrade Controller?
 
-*SUC* is used to assist in the various "Day 2" operations that need to be executed in order to upgrade management/downstream clusters from one Edge platform version to another. Day 2 operations are defined in the form of *SUC Plans*. Based on the these plans, SUC deploys workloads on each node that executes the respective Day 2 operations.
+SUSE Edge uses `SUC` to facilitate various "Day 2" operations related to OS and Kubernetes version upgrades on management and downstream clusters.
+
+"Day 2" operations are defined through `SUC Plans`. Based on these plans, `SUC` deploys workloads on each node to execute the respective "Day 2" operation.
+
+`SUC` is also used within the <<components-upgrade-controller>>. To learn more about the key differences between SUC and the Upgrade Controller, see <<components-upgrade-controller-uc-vs-suc>>.
+
```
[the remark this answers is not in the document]
jiwonhu: Removed the new line at line 27.

asciidoc/components/upgrade-controller.adoc
```diff
@@ -31,6 +33,19 @@ For further details on how the Upgrade Controller works, see <<components-upgrad
...
 
+For information on the difference between the Upgrade Controller and the System Upgrade Controller, see <<components-upgrade-controller-uc-vs-suc>>.
+
```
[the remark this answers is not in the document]
jiwonhu: Reduced the two lines to one line at line 37.

asciidoc/components/upgrade-controller.adoc
```diff
@@ -31,6 +33,19 @@ For further details on how the Upgrade Controller works, see <<components-upgrad
...
+[#components-upgrade-controller-uc-vs-suc]
+== Upgrade Controller vs System Upgrade Controller
+
+The <<components-system-upgrade-controller, System Upgrade Controller (SUC)>> is a general-purpose tool that propagates upgrade instructions to specific Kubernetes nodes.
+
+While it supports some "Day 2" operations for the SUSE Edge platform, it *does not* cover all of them. Moreover, even for supported operations, users have to manually configure, maintain, and deploy multiple `SUC Plans` — an error-prone process that can lead to unexpected issues.
+
+This led to the need for a tool that **automates** and **abstracts** the complexity of managing various "Day 2" operations for the SUSE Edge platform. Thus, the `Upgrade Controller` was developed. It simplifies the upgrade process by introducing a single `user-facing resource` that drives the upgrade. Users only need to manage this resource, while the `Upgrade Controller` takes care of the rest.
+
```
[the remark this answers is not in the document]
jiwonhu: Updated the double new lines into a single one. Thanks!

ipetrov117 (approved):

PR #588, opened 2025-02-20
Updating the support duration for Edge 3.2

rdoxenham:
Confirmed with @timirnich that there's a docs bug with 3.2 set as an 18-month lifecycle, whereas it's 24-months. This PR fixes that.


hardys (approved):

hardys:
> I guess this should also be reflected in our internal `Release Components` document?

Agreed, I have adjusted it now

PR #640, opened 2025-05-07
Edge book versions: Fix typos in image names

Gagrio:
Correcting image names for SL Micro 6.0 raw images to match https://www.suse.com/download/sle-micro/ as that will be the source for customer downloads.

### DIFF
--- a/asciidoc/edge-book/versions.adoc
+++ b/asciidoc/edge-book/versions.adoc
@@ -12,10 +12,10 @@
 :version-edge-registry: 3.2
 
 // == SUSE Linux Micro ==
-:micro-base-image-raw: SL-Micro.x86_64-6.0-Base-GM2.raw
+:micro-base-image-raw: SL-Micro.x86_64-6.0-Base-GM2.raw.xz
 :micro-base-image-iso: SL-Micro.x86_64-6.0-Base-SelfInstall-GM2.install.iso
 :micro-default-image-iso: SL-Micro.x86_64-6.0-Default-SelfInstall-GM2.install.iso
-:micro-default-image-raw: SL-Micro.x86_64-6.0-Default-SelfInstall-GM2.install.raw
+:micro-default-image-raw: SL-Micro.x86_64-6.0-Default-GM2.raw.xz
 :version-sl-micro: 6.0
 
 // == Edge Image Builder ==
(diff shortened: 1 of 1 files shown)


asciidoc/edge-book/versions.adoc
```diff
@@ -12,10 +12,10 @@
 :version-edge-registry: 3.2
 
 // == SUSE Linux Micro ==
-:micro-base-image-raw: SL-Micro.x86_64-6.0-Base-GM2.raw
+:micro-base-image-raw: SL-Micro.x86_64-6.0-Base-GM2.raw.xz
```
hardys, line 15: We can't change this as it's used in the EIB examples, which expect an uncompressed image

asciidoc/edge-book/versions.adoc
```diff
@@ -12,10 +12,10 @@
...
 :micro-base-image-iso: SL-Micro.x86_64-6.0-Base-SelfInstall-GM2.install.iso
 :micro-default-image-iso: SL-Micro.x86_64-6.0-Default-SelfInstall-GM2.install.iso
-:micro-default-image-raw: SL-Micro.x86_64-6.0-Default-SelfInstall-GM2.install.raw
+:micro-default-image-raw: SL-Micro.x86_64-6.0-Default-GM2.raw.xz
```
hardys, line 18: This isn't used anywhere, so lets just remove it to avoid potential confusion when in the majority of cases the Base image is sufficient - sorry for not catching this in the previous review

asciidoc/edge-book/versions.adoc
Gagrio: ah thanks, amending

asciidoc/edge-book/versions.adoc
Gagrio: ack then will clean up shortly

Gagrio:
I made a second commit, as I didn't want to force push (seems to be needed with ammend), but I will squash commits on merge.

hardys:
> I made a second commit, as I didn't want to force push (seems to be needed with ammend), but I will squash commits on merge.

This is fine, but FYI force-push is an acceptable option for PRs, even though in general it's obviously bad practice to rewrite history via force-push

hardys (approved):

PR #644, opened 2025-05-12
add arm64 instructions

alknopfler:
- Add versions for RT kernel (it was missing)
- Add the management cluster configuration in values.yaml metal3 for arm64
- Add the downstream clusters explanation using kiwi in EIB preparation, splitting the current EIB step in different architectures

### DIFF
--- a/asciidoc/product/atip-automated-provision.adoc
+++ b/asciidoc/product/atip-automated-provision.adoc
@@ -340,18 +340,22 @@ Much of the configuration is possible with Edge Image Builder, but in this guide
 ==== Prerequisites for air-gap scenarios
 
 * A container runtime such as https://podman.io[Podman] or https://rancherdesktop.io[Rancher Desktop] is required to run Edge Image Builder.
-* For the x86_64 architecture, the base image `{micro-base-image-raw}` (or `{micro-base-rt-image-raw}` for the Real-Time kernel) must be downloaded from the https://scc.suse.com/[SUSE Customer Center] or the https://www.suse.com/download/sle-micro/[SUSE Download page].
-* For the arm64 architecture, the base image will be built using the following guide <<guides-kiwi-builder-images>> with the profile `Base-SelfInstall` (or `Base-RT-SelfInstall` for the Real-Time kernel).
+* The base image will be built using the following guide <<guides-kiwi-builder-images>> with the profile `Base-SelfInstall` (or `Base-RT-SelfInstall` for the Real-Time kernel). The process is the same for both architectures (x86-64 and aarch64).
+* To deploy aarch64 downstream clusters, you must set before the management cluster deployment the `deployArchitecture: arm64` parameter in the `metal3.yaml` file explained in <<arm64-mgmt-cluster,Management Cluster Documentation>>. This is required to ensure that the correct architecture is used for the downstream cluster.
 * If you want to use SR-IOV or any other workload which require a container image, a local private registry must be deployed and already configured (with/without TLS and/or authentication). This registry will be used to store the images and the helm chart OCI images.
+
+[NOTE]
+====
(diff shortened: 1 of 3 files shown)


asciidoc/product/atip-automated-provision.adoc
```diff
@@ -67,14 +67,17 @@ Much of the configuration via Edge Image Builder is possible, but in this guide,
 ==== Prerequisites for connected scenarios
 
 * A container runtime such as https://podman.io[Podman] or https://rancherdesktop.io[Rancher Desktop] is required to run Edge Image Builder.
-* The base image `{micro-base-image-raw}` must be downloaded from the https://scc.suse.com/[SUSE Customer Center] or the https://www.suse.com/download/sle-micro/[SUSE Download page].
+* For the x86_64 architecture, the base image `{micro-base-image-raw}` (or `{micro-base-rt-image-raw}` for the Real-Time kernel) must be downloaded from the https://scc.suse.com/[SUSE Customer Center] or the https://www.suse.com/download/sle-micro/[SUSE Download page].
+* For the arm64 architecture, the base image will be built using the following guide <<guides-kiwi-builder-images>> with the profile `Base-SelfInstall` (or `Base-RT-SelfInstall` for the Real-Time kernel).
```
hardys, line 71: I think we'll need to use kiwi to build the image in all cases because the 6.0/6.1 GM images don't contain the fix for https://bugzilla.suse.com/show_bug.cgi?id=1240375 ?

alknopfler: you mean for ARM? because x86_64 we don't need the fix about UEFI

alknopfler: if your comment is related to 6.0 and 6.1 I think with the templating is enough because the process will build for both images (different release docs)

alknopfler: you are right... let me change it... I forgot the other issue with the nginx

alknopfler: changed!

asciidoc/product/atip-management-cluster.adoc
```diff
@@ -839,7 +839,31 @@ metal3-ironic:
...
+    predictableNicNames: "true"
+  persistence:
+    ironic:
+      size: "5Gi"
+----
+
+[NOTE]
+====
+In the current version, a limitation exists regarding the use of `deployArchitecture: arm64`. Specifically, if you enable the deployment of downstream arm64 clusters using this directive, the management cluster will subsequently only be able to deploy this architecture.
+To deploy clusters on both architectures (x86_64 and arm64), you will need to provision two separate management clusters. This limitation will be removed in future versions.
```
hardys, line 865: nit - I'd say "removed in a future version"

hardys (dismissed):

asciidoc/product/atip-management-cluster.adoc
alknopfler: changed! Thanks

asciidoc/product/atip-automated-provision.adoc
```diff
@@ -1431,6 +1450,7 @@ The directed network provisioning workflow allows to automate the Telco features
 
 *Requirements*
 
+- The image generated using `EIB` has to include the base images downloaded from https://scc.suse.com/[SUSE Customer Center] or the https://www.suse.com/download/sle-micro/[SUSE Download page] with the RT kernel `{micro-base-rt-image-raw}`.
```
fdegir, line 1453: I might be missing the context here but I'm a bit confused about this requirement as the Image Configuration chapters above refer to Kiwi built base images while this requirement states it should be the base image downloaded from SCC. Isn't this requirement in conflict with the previous paragraphs?

alknopfler: you're totally right! good catch Fatih... The problem is that the file is toooo long so I missed the last one... 
Thanks again good catch!

fdegir (approved):
LGTM

hardys (approved):
lgtm!

PR #647, opened 2025-05-15
[3.1.2] - update atip mgmt cluster airgap images

dprodanov4:


hardys:
Looks good but I think we need to make similar updates to `guides/air-gapped-eib-deployments.adoc` too?

asciidoc/edge-book/releasenotes.adoc
```diff
@@ -150,7 +150,7 @@ Summary: SUSE Edge 3.1.1 is the first release z-stream in the SUSE Edge 3.1 rele
 
 == New Features
 
-* The NeuVector version is updated to `5.4.0` which provides several new features: https://open-docs.neuvector.com/releasenotes/5x#release-notes-for-5x[Release Notes]
+* The NeuVector version is updated to `5.4.2` which provides several new features: https://open-docs.neuvector.com/releasenotes/5x#release-notes-for-5x[Release Notes]
```
hardys, line 153: I think this file update was a mistake?  We shouldn't adjust any versions for previous releases in the releasenotes

dprodanov4: nice catch, its fixed now

PR #656, opened 2025-05-20
Document the NVMe queue allocation with managed_irq

mchiappero:
Document the behavior of "managed_irq" in Edge 3.3 with respect to the IRQ lines for NVMe devices, which are managed by the kernel. In particular all the queues are allocated to housekeeping cores, and limited to the amount of such cores.


hardys (approved):
lgtm - are additional changes required before moving this out of draft status?

hardys (approved):

PR #721, opened 2025-06-26
Added a note on how to incorporate feedback on a PR

e-minguez:


e-minguez:
There are multiple ways to do this, this is how _I_ do it but open to suggestions to better ways :)

mchiappero:
> Usually with a single commit I do `git commit --amend` and with multiple ones `git commit --fixup=commit-to-fix` then a `git rebase -i --autosquash` before push force. I find it easier than the reset solution (I only use the reset solution when I need to split a commit in fact), as you can easily see the differences with your previous submission before modifying the commits.
> 
> But I think you'll likely end up with a different solution per person you ask the question to.

No, this is fundamentally the right and generic way. The soft reset is necessary for moving files while reworking the patchset, which should be done anyway with an interactive rebase.

e-minguez:
I've fixed the oneline thing. Can I get a review? I really want this to be merged :) We can of course improve it later but at least we have _something_

Issue #688, opened 2025-06-17
[doc] Issue in "Management cluster definition file" - Package git required?


peteingithub:
In the full YAML file packages listed are `git` and `jq`. In the explanatory section below, only `jq` is included.
I recall `git` being required for baremetal deployment, but is it required for the Mgmt cluster.

PR #765, opened 2025-07-24
remove git package from mgmt

alknopfler:
- fix to remove git package


alknopfler:
fix issue: https://github.com/suse-edge/suse-edge.github.io/issues/688

Issue #472, opened 2024-10-15
[doc] Issue in "Management cluster definition file" - Longhorn problem


peteingithub:
Management cluster definition file:

https://documentation.suse.com/suse-edge/3.1/html/edge/atip-management-cluster.html#mgmt-cluster-image-definition-file

Longhorn requires `open-iscsi`. 
Does that belong to the SLM6 Base ISO? 
Should Default ISO be used?
Or some other solution?

hardys:
@peteingithub thanks for the report - I think the solution is to add open-iscsi to the EIB `packageList` - in Edge 3.0 we used the "Default" image, but in 3.1 we switched to the "Base" image which evidently does not contain this by default

alknopfler:
For the mgmt-cluster as you don't need RT kernel, you can also use the default image as base image to avoid installing the open-iscsi package.

We will update the examples and docs to make it clear
Thanks Petr

andreacasini:
The issue is still present today, 3.3 docs instruct to use "SL-Micro.x86_64-6.1-Base-SelfInstall-GM.install.iso" but then in the "packageList" section of the "mgmt-cluster.yaml" definition file the open-iscsi package is still missing and this stops the mgmt-cluster to install properly

PR #768, opened 2025-07-30
added open-iscsi to packagelist to allow longhorn to install

andreacasini:
In the `40.3.2 Management cluster definition file` section, the example file `mgmt-cluster.yaml` is listing longhorn and metal3 helm charts among the others to get installed.

Previously, the `SL-Micro.x86_64-6.1-Base-SelfInstall-GM.install.iso` image file is recommended to be used.

The package `open-iscsi` is a hard requirement for longhorn and such package is not included in the `SL-Micro.x86_64-6.1-Base-SelfInstall-GM.install.iso` image, which leads to a condition where longhorn is never completing the install process also impacting metal3 install process because it is waiting for its PVC to be provisioned.

It's necessary to add `open-iscsi` to the packageList in order for setup to successfully complete.

Fixes: #472

### DIFF
--- a/asciidoc/product/atip-management-cluster.adoc
+++ b/asciidoc/product/atip-management-cluster.adoc
@@ -158,6 +158,7 @@ operatingSystem:
   packages:
     packageList:
     - jq
+    - open-iscsi
     sccRegistrationCode: $SCC_REGISTRATION_CODE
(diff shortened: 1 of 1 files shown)


hardys (approved):
Thanks!

PR #792, opened 2025-08-18
Drop kube-controller-manager workaround

fdegir:
It's been identified that the use of EIB Kubernetes manifests mechanism to update the HelmChartConfigs for the core Helm charts results in race condition, causing the Kubernetes kube-controller-manager to throw an unhandled exception due to both RKE2 HelmController and kube-controller-manager to operate on the same resource.

To avoid triggering the issue, EIB documentation must be updated to instruct users **not** to use EIB Kubernetes manifests and instead use EIB **os-files**.

This PR is proposed to ensure the workaround can be dropped from main branch and backported to 3.3 for the upcoming maintenance release as use of os-files is successfully validated.


hardys:
Pushed https://github.com/suse-edge/suse-edge.github.io/pull/802 which adjusts what I think is the only relevant example in the docs, so we can merge that, then this PR and backport both to 3.3

hardys (approved):

PR #799, opened 2025-08-20
[3.3.2] - update docs for 3.3.2 release

dprodanov4:


hardys:
> The changes for `os-files` are now covered under this [PR](https://github.com/suse-edge/edge-image-builder/pull/764/files)

Users may not look in the EIB upstream docs, I think we need to copy similar content into the release notes to explain that it fixes one of the known issues in 3.3.1 - I can do this in my follow-up PR if you like 👍 

As discussed on Slack we also need to update the airgap images to match the new versions, but otherwise lgtm, thanks!

Issue #829, opened 2025-09-15
Upgrade table/matrix to clarify the upgrade procedures


e-minguez:
For downstream clusters, we have (at least) 3 different scenarios, and each of them involve different upgrade procedures/mechanisms. AFAIK:

* For EIB provisioned clusters -> Fleet approach
* For Metal3 provisioned clusters -> Redeploy with a fresh image
* For phone-home provisioned clusters -> K8s upgrade via rancher + SUC for the OS and everything else

It would be nice to have a table/matrix/something to explicitly clarify each scenario.

cc @hardys 

@ranjinimn is this something you can take care of?

ranjinimn:
@e-minguez I can work on this. Is this all the information that you want to include in the matrix? Any specific section where this needs to be included? Would https://documentation.suse.com/suse-edge/3.3/single-html/edge/#day-2-operations be okay?

e-minguez:
> [@e-minguez](https://github.com/e-minguez) I can work on this. Is this all the information that you want to include in the matrix? 

I think so, what do you think?

> Any specific section where this needs to be included? Would https://documentation.suse.com/suse-edge/3.3/single-html/edge/#day-2-operations be okay?

Yes.

ranjinimn:
@e-minguez I was wondering if we can cross reference the respective sections which details the steps provided for upgrade. For example, you mention the Fleet approach, can we just provide the link to that section which mentions about upgrading using Fleet?

e-minguez:
> [@e-minguez](https://github.com/e-minguez) I was wondering if we can cross reference the respective sections which details the steps provided for upgrade. For example, you mention the Fleet approach, can we just provide the link to that section which mentions about upgrading using Fleet?

That would be nice.

PR #859, opened 2025-09-24
Added the table for clusters and methods for upgrade

ranjinimn:
https://github.com/suse-edge/suse-edge.github.io/issues/829

I have created the table as per the description provided. 

For downstream clusters, we have (at least) 3 different scenarios, and each of them involve different upgrade procedures/mechanisms. AFAIK:

For EIB provisioned clusters -> Fleet approach
For Metal3 provisioned clusters -> Redeploy with a fresh image
For phone-home provisioned clusters -> K8s upgrade via rancher + SUC for the OS and everything else
It would be nice to have a table/matrix/something to explicitly clarify each scenario.

### DIFF
--- a/asciidoc/day2/migration.adoc
+++ b/asciidoc/day2/migration.adoc
@@ -1,6 +1,6 @@
 [#day2-migration]
 = Edge 3.3 migration
-:revdate: 2025-05-20
+:revdate: 2025-09-24
 :page-revdate: {revdate}
 :experimental:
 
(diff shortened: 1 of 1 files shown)


e-minguez:
> @hardys and @e-minguez I created the table based on the information provided. One question I had was regarding the inclusion of the "Upgrade controller" in this table? Is there any specific cluster for which we need to use this? Or is this generic?

https://documentation.suse.com/suse-edge/3.4/html/edge/components-upgrade-controller.html#components-upgrade-controller-known-issues

> Downstream cluster upgrades are not yet managed by the Upgrade Controller.
Also
> Currently the Upgrade Controller supports non air-gapped environment upgrades only. Air-gapped upgrades are not yet possible.

e-minguez (approved):

e-minguez (approved):

PR #908, opened 2025-10-15
Cherry-pick-859-3.1 version - Added the table for clusters and methods for upgrade (#859)

ranjinimn:
https://github.com/suse-edge/suse-edge.github.io/pull/859/files

https://github.com/suse-edge/suse-edge.github.io/issues/829#issuecomment-3405605118

PR #914, opened 2025-10-16
[release-3.3] workflows: enable running on release-3.* PRs

hardys:
Backport #911

Currently issues like that mentioned in #908 won't be caught because we're not testing stable branch PRs

(cherry picked from commit 7229178b18868260686a7526e1c38f5fb80f1b35)


e-minguez:
> @e-minguez added a fix commit related to #890, seems we missed a file in that backport which results in a broken build

Thank you so much!

Issue #924, opened 2025-10-22
[doc] Issue in "Install PTP software components" - Old apiVersion


peteingithub:
Install PTP software components:

https://documentation.suse.com/suse-edge/3.4/single-html/edge/#id-install-ptp-software-components

apiVersion: 1.0 -> apiVersion: 1.3

PR #935, opened 2025-10-30
atip-features: fix templating and hard-coded version

hardys:
The EIB apiVersion should be set via a variable, and we're missing the attributes configuration so the micro-base-rt-image-raw is not correctly substituted.

Fixes: #924


hardys:
> The changes looks good. Just a question here. What about the base image -"{micro-base-rt-image-raw}" . Is it meant to be that way in the yaml file?

Yes similar to other locations this should specify the micro base image, but due to the missing attributes configuration we can see it's not currently working ref https://suse-edge.github.io/atip-features.html#id-install-ptp-software-components

PR #936, opened 2025-10-31
WIP : Add CI check for broken links

hardys:


hardys:
@ranjinimn @fsundermeyer - from local testing I don't think linkspector will work for this as it doesn't consider URLs which contain attributes - can you offer any guidance re alternative ways to catch broken links?

hardys:
@fsundermeyer thanks a lot for the insights - the `linkspector` tool is https://github.com/UmbrellaDocs/linkspector - I found it while searching for github actions which can check links in asciidoc - https://github.com/UmbrellaDocs/action-linkspector

Not using attributes for links may be workable - if we keep all the links in one file we can use sed to replace all-the-versions when bumping, my main concern is people may forget (particularly when doing minor version updates for z-stream releases), perhaps it's possible to write a small script to check though.

I am also aware that our check job is using an old daps image built by @agracey ref https://github.com/suse-edge/suse-edge.github.io/blob/main/.github/workflows/test-deploy.yml#L26 - I don't recall exactly why we needed that but perhaps it's a good time to review and decide if we can align better with how other projects are tested, any guidance on that?

PR #939, opened 2025-11-04
fix: frr-k8s versions for 3.4

e-minguez:
```
❯ crane ls -O registry.suse.com/edge/3.4/frr-k8s
v0.0.16
v0.0.16-1.1
v0.0.16-1.7
```


hardys:
AFAICS it's not only the frr-k8s image which is wrong, so I pushed https://github.com/suse-edge/suse-edge.github.io/pull/941 as an alternative (or feel free to copy those changes here if you prefer 👍 )

PR #947, opened 2025-11-05
ISSUE#620 -Adding control-plane nodes" - Requires tls-sans to be set for LB

### LINKED ISSUE #620: [doc] Issue in "Adding control-plane nodes" - Requires tls-sans to be set for LB
```
# Export the VIP_SERVICE_IP in the VM
# Replace with the actual IP
export VIP_SERVICE_IP=<ip>

curl -sfL https://get.rke2.io | INSTALL_RKE2_TYPE="server" sh -
systemctl enable rke2-server.service
```
Add tls-sans to /etc/rancher/rke2/config.yaml before starting the service to make sure that kubectl works in load-balanced environment.

ranjinimn:
Closes #620 

 Edited content based on the information in the issue

### DIFF
--- a/asciidoc/guides/metallb-kube-api.adoc
+++ b/asciidoc/guides/metallb-kube-api.adoc
@@ -289,21 +289,29 @@ Now execute the commands below on the second and third nodes.
 For RKE2:
 [,bash]
 ----
-# Export the VIP_SERVICE_IP in the VM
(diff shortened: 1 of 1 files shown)


ranjinimn:
@e-minguez Can you please let me know to what all version do I need to port this content?

PR #957, opened 2025-11-06
Cherry-pick-3.4-620-Adding control-plane nodes" - Requires tls-sans to be set for LB

### LINKED ISSUE #620: [doc] Issue in "Adding control-plane nodes" - Requires tls-sans to be set for LB

ranjinimn:
Closes #620 

Edited content based on the information in the issue (#947)

### DIFF
--- a/asciidoc/guides/metallb-kube-api.adoc
+++ b/asciidoc/guides/metallb-kube-api.adoc
@@ -292,7 +292,7 @@ For RKE2:
 # As a root user, create the /etc/rancher/rke2/config.yaml config file with the following content:
 
 mkdir -p /etc/rancher/rke2/
-cat << EOF > /etc/rancher/rke2/config.yaml
+cat <<EOF > /etc/rancher/rke2/config.yaml
 # An example of the config.yaml file for an agent node: 
 server: https://${VIP_SERVICE_IP}:9345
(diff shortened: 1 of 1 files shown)


e-minguez:
> @e-minguez is the EOF space issue shell specific? I just tried it on zsh and bash and it works fine with or without the space?

I promised I tested it and it didn't worked... now I've tested it again with zsh and bash and it works. Apologies for the noise (again)

hardys:
> > @e-minguez is the EOF space issue shell specific? I just tried it on zsh and bash and it works fine with or without the space?
> 
> I promised I tested it and it didn't worked... now I've tested it again with zsh and bash and it works. Apologies for the noise (again)

Ok, definitely good to check these things - as e.g leading space for the closing EOF will cause problems, but if we're agreed that in this case it's cosmetic lets merge the backports and not block on the change to main, thanks!

PR #962, opened 2025-11-13
fix: Add a note about checking image checksum

e-minguez:


ranjinimn (commented):
Just had a small suggestion

Issue #896, opened 2025-10-10
[doc] "Standalone clusters with Edge Image Builder" doesn't explain the os-files feature


e-minguez:
https://documentation.suse.com/suse-edge/3.4/html/edge/quickstart-eib.html#

https://github.com/suse-edge/edge-image-builder/blob/main/docs/building-images.md#operating-system-files

hardys:
@ranjinimn I think this can be closed now https://github.com/suse-edge/suse-edge.github.io/pull/909 merged?

Note that when linking a PR to an issue, if you use one of the special keywords in the PR description e.g "fixes #nnn" the relevant issue will auto-close https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

hardys:
> I shall close this once I backport the content. Is this relevant for all the earlier versions [@hardys](https://github.com/hardys) ?

Sorry @ranjinimn missed this question - yes this is relevant to all currently supported Edge releases so it would be good to backport if possible

PR #909, opened 2025-10-15
Added content on Operating System Files

ranjinimn:
Closes #896

PR #965, opened 2025-11-14
Cherry-pick-3.2-896 - [doc] "Standalone clusters with Edge Image Builder" doesn't explain the os-files feature

ranjinimn:
Closes https://github.com/suse-edge/suse-edge.github.io/issues/896

Added content on Operating System Files (https://github.com/suse-edge/suse-edge.github.io/pull/909)

@hardys 2 new sections have been added. Please review.


hardys:
> I think it would be better if the PR was two commits, one per change instead a single one, but I'll let @hardys to chime in :)

Sorry for the slow reply - yes I agree this is also backporting https://github.com/suse-edge/suse-edge.github.io/commit/7f9eb5237ddea0f06711191d409f15a4c8fa0be1 which was merged via #725 

If we want to backport both changes that's fine, but it should be cherry-picked as two commits, and both changes should be mentioned in the PR description

Alternatively we can remove the certificates part and propose that via a follow-up PR?

hardys (changes requested):
As discussed please either split into two commits or remove the changes to #725 and propose via a follow-up PR, thanks!

ranjinimn:
@e-minguez I have added a new commit to this PR. please review.

PR #976, opened 2025-11-20
Enhance vRAN Acceleration and CPU Pinning chapters

jiwonhu:
This merge request aims to restructure and consolidate telco feature configurations. It is to align the documentation with complete end-user requirements (e.g. CPU pinning), rather than disparate technologies (e.g. kubelet config).

CPU Pinning
* merge all previously scattered config steps (e.g. kernel arguments, TuneD, kubelet settings) into a single, unified chapter to provide a complete configuration picture. Also, improve readability by restructuring large blocks of text into clear subsections
* Add instructions for kubelet config file for recent RKE2 versions. Kubelet arguments no longer work

vRAN Acceleration
* Add explicit mention of the dependency on SRIOV within vRAN acceleration chapter, ensuring users know the full steps for configuration.
* Elaborate on card models (ACC100/vRAN Boost) and their form factors
* Update vRAN accelerator VF count from 1 to higher number

### DIFF
--- a/asciidoc/product/atip-features.adoc
+++ b/asciidoc/product/atip-features.adoc
@@ -21,7 +21,8 @@ The following topics are covered in this section:
 
 * <<kernel-image-for-real-time,Kernel image for real time>>: Kernel image to be used by the real-time kernel.
 * <<kernel-args,Kernel arguments for low latency and high performance>>: Kernel arguments to be used by the real-time kernel for maximum performance and low latency running telco workloads.
-* <<cpu-pinning,CPU Pinning>>: Isolating the CPUs via kernel arguments, TuneD profile, and Kubernetes configuration.
+* <<cpu-pinning-host,CPU Pinning on Host>>: Isolating the CPUs via kernel arguments and TuneD profile.
+* <<cpu-pinning-k8s,CPU Pinning on Kubernetes>>: Isolating the CPUs on Kubernetes via Kubelet configuration.
 * <<cni-configuration,CNI configuration>>: CNI configuration to be used by the Kubernetes cluster.
 * <<sriov,SR-IOV configuration>>: SR-IOV configuration to be used by the Kubernetes workloads.
 * <<dpdk,DPDK configuration>>: DPDK configuration to be used by the system.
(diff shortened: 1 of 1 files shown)


jiwonhu:
Would be good if @alknopfler and @antaloala could review. As this PR is influenced by our recent customer project

asciidoc/product/atip-features.adoc
```diff
@@ -21,13 +21,12 @@ The following topics are covered in this section:
 
 * <<kernel-image-for-real-time,Kernel image for real time>>: Kernel image to be used by the real-time kernel.
 * <<kernel-args,Kernel arguments for low latency and high performance>>: Kernel arguments to be used by the real-time kernel for maximum performance and low latency running telco workloads.
-* <<cpu-tuned-configuration,CPU Pinning via Tuned and kernel args>>: Isolating the CPUs via kernel arguments and Tuned profile.
+* <<cpu-pinning,CPU Pinning>>: Isolating the CPUs via kernel arguments, TuneD profile, and Kubernetes configuration.
```
alknopfler: I can understand your idea, but our decision some time ago was to keep both sections separately, because you can tune the system without modifying the RKE2 part for some uses cases where you don't need to isolate RKE2 parts.
What we can do is to keep both separately but closer to each other (first cpu pinning and then cpu-pinning-kubernetes.
WDYT?

asciidoc/product/atip-features.adoc
```diff
@@ -1123,13 +1234,16 @@ Network devices using kernel driver
...
 
-One of the most compute-intensive 4 G and 5 G workloads is RAN layer 1 (`L1`) `FEC`, which resolves data transmission errors over unreliable or noisy communication channels. `FEC` technology detects and corrects a limited number of errors in 4 G or 5 G data, eliminating the need for retransmission. Since the `FEC` acceleration transaction does not contain cell state information, it can be easily virtualized, enabling pooling benefits and easy cell migration.
+One of the most compute-intensive 4G and 5G workloads is RAN layer 1 (L1) `FEC`, which resolves data transmission errors over unreliable or noisy communication channels. `FEC` technology detects and corrects a limited number of errors in 4G or 5G data, eliminating the need for retransmission. Since the `FEC` acceleration transaction does not contain cell state information, it can be easily virtualized, enabling pooling benefits and easy cell migration.
```
alknopfler: IIRC all acronymous should be under quotes (format style) @ranjinimn . Am I right?
```suggestion
One of the most compute-intensive 4G and 5G workloads is RAN layer 1 (`L1`) `FEC`, which resolves data transmission errors over unreliable or noisy communication channels. `FEC` technology detects and corrects a limited number of errors in 4G or 5G data, eliminating the need for retransmission. Since the `FEC` acceleration transaction does not contain cell state information, it can be easily virtualized, enabling pooling benefits and easy cell migration.
```

asciidoc/product/atip-features.adoc
```diff
@@ -1191,19 +1307,20 @@ $ dpdk-devbind.py -b vfio-pci 0000:8a:00.0
...
 ----
-$ echo 2 > /sys/bus/pci/devices/0000:8a:00.0/sriov_numvfs
-$ dpdk-devbind.py -b vfio-pci 0000:8b:00.0
+$ cat /sys/bus/pci/devices/0000:8a:00.0/sriov_totalvfs
+16
+$ echo 16 > /sys/bus/pci/devices/0000:8a:00.0/sriov_numvfs
 ----
 
-* Configure acc100 with the proposed configuration file:
+* Configure the acclerator card and its virtual functions with a 5G profile. A unique VF token (UUID) must be provided. The workload would consume this VF token to utilize the card's FEC acceleration capabilities:
```
alknopfler: We cannot force to a specific profile. For LTE they should use the 4G for example. I'd explain that inside the /opt/pf-bb-config there are some options to be used (there are many profiles inside to be used, so this is only an example IMO)
```suggestion
* Configure the acclerator card and its virtual functions with a 4G or 5G profile. A unique VF token (UUID) must be provided. The workload would consume this VF token to utilize the card's FEC acceleration capabilities:
```

asciidoc/product/atip-features.adoc
```diff
@@ -1191,19 +1307,20 @@ $ dpdk-devbind.py -b vfio-pci 0000:8a:00.0
...
 
 [,shell]
 ----
-$ pf_bb_config ACC100 -c /opt/pf-bb-config/acc100_config_vf_5g.cfg
+$ pf_bb_config ACC100 -c /opt/pf-bb-config/acc100_config_vf_5g.cfg -v 00112233-4455-6677-8899-aabbccddeeff
```
alknopfler: I think we should add the ACC200 example we've got too here... WDYT? 
Also, I think we should explain the token part having to possibility to improve this section...

alknopfler (commented):
First of all, thanks for your help improving this section!!! 

Some comments ;) 
Thanks again

PR #762, opened 2025-07-22
[SEL3-41] Documenting ironic known issue for network device naming

Gagrio:
https://jira.suse.com/projects/SEL3/issues/SEL3-41?filter=allopenissues

PR #994, opened 2025-12-12
Revert note about Ironic known issue

diconico07:
As both IPA and SL Micro are now using the same code stream, all components that may trigger this behavior are aligned. So this is no longer an issue.

This reverts PR #762 .


e-minguez (approved):
I'm all in for removing known issues! LGTM

PR #996, opened 2025-12-15
rancher-turtles: Updates for 2.13

hardys:
This adds updates to align with the latest Rancher version which enables turtles by default.

It also aligns with the new Edge rancher-turtles-providers chart added to factory ref https://src.opensuse.org/suse-edge/Factory/src/branch/main/rancher-turtles-providers-chart - note that airgap scenarios will require https://src.opensuse.org/suse-edge/Factory/pulls/331 which is not yet merged


dbw7 (approved):
LGTM

PR #1008, opened 2026-01-08
Add script to generate versions.adoc

hardys:
Adds a script/template which allows us to generate the versions derived from the release manifest, hopefully this will be less effort and less error-prone than manual updates


scripts/versions_update.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+
+
+set -eux
+
+RELEASE_BRANCH=${RELEASE_BRANCH_:-"main"}
+
+echo "Downloading release manifest for ${RELEASE_BRANCH} branch"
+curl "https://src.opensuse.org/suse-edge/Factory/raw/branch/${RELEASE_BRANCH}/release-manifest-image/release_manifest.yaml" -o _release_manifest.yaml
```
diconico07: Would rather use the rendered version from `registry.opensuse.org` here, just to avoid the templating issues

asciidoc/edge-book/versions.adoc.j2
```diff
@@ -0,0 +1,154 @@
...
+:version-longhorn-chart: {{ release_manifest_longhorn_version }}
+:version-longhorn-crd-chart: {{ release_manifest_longhorn_crd_version }}
+:version-longhorn-docs: {{ release_manifest_longhorn_version.split("+up") | last }}
+:version-metal3-chart: {{ release_manifest_metal3_version }}
+:version-metallb-chart: {{ release_manifest_metallb_version }}
+:version-neuvector-chart: {{ release_manifest_neuvector_version }}
+:version-neuvector-crd-chart: {{ release_manifest_neuvector_crd_version }}
+:version-neuvector-dashboard-extension-chart: {{ release_manifest_neuvector_dashboard_version }}
+:version-rancher-chart: {{ release_manifest_rancher_version }}
+:version-rancher-turtles-chart: 304.0.6+up0.24.0
```
diconico07: I guess this one should just get replaced with the turtles providers chart.

diconico07 (commented):
Mostly usability comments. This will be a really nice addition 👍

asciidoc/edge-book/versions.adoc.j2
hardys: Yes, but I was planning to handle that in https://github.com/suse-edge/suse-edge.github.io/pull/996

scripts/versions_update.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
hardys: Yeah fair point, I can re-write it in python which will probably be cleaner

scripts/versions_update.py
```diff
@@ -0,0 +1,288 @@
...
+
+    kiwi_builder_repo = f"{registry_base}/kiwi-builder"
+    kiwi_builder_version = get_latest_version_tag(kiwi_builder_repo)
+    # Strip suffix after hyphen (e.g., 10.2.29.1-1.1 -> 10.2.29.1)
+    manifest_data['version_kiwi_builder'] = kiwi_builder_version.split('-')[0]
+
+    eib_repo = f"{registry_base}/edge-image-builder"
+    eib_version = get_latest_version_tag(eib_repo)
+    # Strip suffix after hyphen (e.g., 1.3.3-4.4 -> 1.3.3)
+    manifest_data['version_eib'] = eib_version.split('-')[0]
```
diconico07: Those two are now in the tooling manifest

scripts/versions_update.py
```diff
@@ -0,0 +1,288 @@
...
+    release_version = manifest_data.get('spec', {}).get('releaseVersion', '')
+    if release_version:
+        version_parts = release_version.split('.')[:2]  # Get major.minor
+        chart_prefix = version_parts[0] + '0' + version_parts[1]  # e.g., 3.5 -> 305
+        # Charts are at registry.suse.com/edge/charts/upgrade-controller (not version-specific)
+        registry_name = repository.split('/')[0]
+        upgrade_controller_repo = f"{registry_name}/edge/charts/upgrade-controller"
+        upgrade_controller_version = get_latest_version_tag_with_prefix(upgrade_controller_repo, chart_prefix)
+        # Strip suffix after hyphen (e.g., 305.0.3_up0.1.3-4.1 -> 305.0.3_up0.1.3)
+        manifest_data['version_upgrade_controller_chart'] = upgrade_controller_version.split('-')[0]
```
diconico07: Upgrade controller is now in release manifest

scripts/versions_update.py
```diff
@@ -0,0 +1,288 @@
...
+    template_file = Path(template_path)
+
+    # Load template
+    env = Environment(loader=FileSystemLoader(template_file.parent))
+    template = env.get_template(template_file.name)
+
+    # Render template with manifest data
+    output = template.render(manifest_data)
+
+    # Replace %%REVDATE%% with current date
```
diconico07: Why not add current date to the variables available to the template ?

asciidoc/edge-book/versions.adoc.j2
```diff
@@ -0,0 +1,147 @@
...
+{%- set release_manifest_chart_major = release_manifest_edge_version_short | replace(".", "0") %}
+{%- set release_manifest_metallb_version = spec.components.workloads.helm | selectattr("releaseName", "equalto", "metallb") | map(attribute="version") | list | first  %}
+{%- set release_manifest_eco_version = spec.components.workloads.helm | selectattr("releaseName", "equalto", "endpoint-copier-operator") | map(attribute="version") | list | first  %}
+{%- set release_manifest_cdi_version = spec.components.workloads.helm | selectattr("releaseName", "equalto", "cdi") | map(attribute="version") | list | first  %}
+{%- set release_manifest_akri_version = spec.components.workloads.helm | selectattr("releaseName", "equalto", "akri") | map(attribute="version") | list | first  %}
+{%- set release_manifest_akri_dashboard_version = spec.components.workloads.helm | selectattr("releaseName", "equalto", "akri") | map(attribute="addonCharts") | first | map(attribute="version") | list | first  %}
```
diconico07: These two are no more

diconico07 (changes requested):

scripts/versions_update.py
```diff
@@ -0,0 +1,374 @@
...
+        fleet_version = get_image_version_from_rancher(rancher_version, 'rancher/fleet')
+        manifest_data['version_fleet'] = fleet_version
+
+        suc_chart_version = get_suc_chart_version_from_rancher(rancher_version)
+        manifest_data['version_suc_chart'] = suc_chart_version
+    else:
+        print("⚠ Warning: Rancher version not found in manifest, skipping SUC and Fleet versions")
+
+    # Define revision_date
+    manifest_data['revision_date'] = date.today().strftime('%Y-%m-%d')
```
diconico07: nit: for reproducibility, we may want to use the build date of the release manifest we are fetching, wdyt ? (i.e. parsing the `org.opencontainers.image.created` label)

diconico07 (dismissed):
LGTM

scripts/versions_update.py
hardys: done, thanks!

scripts/versions_update.py
hardys: Thanks, that's a good idea - I was also thinking about how we might trigger a renovate workflow in future - I guess we could use this label instead of just watching the tag to ensure we always update even when the tag isn't bumped 🤔 

It might also be nice to write the container tag, revision and hash as a comment at the start of the rendered file just to make it completely clear where the data is derived from

hardys: now done

hardys:
@diconico07 this is ready for another review when you get time - note I switched to the OBS release-manifest and added a GH workflow which aims to regularly update when it changes - if that works OK we can do the same for release branches but adjust the `--image` argument to reference the product manifests

asciidoc/edge-book/versions.adoc.j2
```diff
@@ -0,0 +1,144 @@
...
+{%- set release_manifest_sriov_version = spec.components.workloads.helm | selectattr("releaseName", "equalto", "sriov-network-operator") | map(attribute="version") | list | first %}
+{%- set release_manifest_sriov_crd_version = spec.components.workloads.helm | selectattr("releaseName", "equalto", "sriov-network-operator") | map(attribute="dependencyCharts") | first | map(attribute="version") | list | first %}
+{%- set release_manifest_turtles_providers_version = spec.components.workloads.helm | selectattr("releaseName", "equalto", "rancher-turtles-providers") | map(attribute="version") | list | first %}
+{%- set release_manifest_cert_manager_version = spec.components.workloads.helm | selectattr("releaseName", "equalto", "cert-manager") | map(attribute="version") | list | first %}
+{%- set release_manifest_upgrade_controller_version = spec.components.workloads.helm | selectattr("releaseName", "equalto", "upgrade-controller") | map(attribute="version") | list | first %}
+{%- set release_manifest_private_registry_version = spec.components.workloads.helm | selectattr("releaseName", "equalto", "private-registry-helm") | map(attribute="version") | list | first %}
```
hardys, line 29: One open question is whether we should keep this templating to extract the release-manifest data or just extract it in the python script similar to https://github.com/hardys/fleet-examples/blob/templating/scripts/fleet_versions_update.py#L257

hardys: I think perhaps leaving it like this may be best, then it's simple to add new charts without any need to modify the script - it could be argued that the script approach would be cleaner though.

asciidoc/edge-book/versions.adoc
```diff
@@ -10,47 +13,46 @@
...
+:version-rancher-prime: 2.14.1
+:version-cert-manager: 1.20.1
+:version-elemental-operator: 1.9.0
+:version-longhorn: 1.11.1
+:version-suc: v0.19.1
+:version-fleet: v0.15.1
 :version-nvidia-device-plugin: 0.14.5
 :version-kiwi-builder: 10.2.29.1
 :version-private-registry: 1.1.1
-:version-ipa: 3.0.8
```
diconico07, line 39: This one is new and should stay

asciidoc/edge-book/versions.adoc
```diff
@@ -60,24 +62,25 @@
...
-:version-neuvector-crd-chart: 108.0.1+up2.8.10
-:version-rancher-turtles-providers-chart: 305.0.4+up0.25.1
-:version-sriov-crd-chart: 305.0.4+up1.6.0
-:version-sriov-network-operator-chart: 305.0.4+up1.6.0
+:version-cdi-chart: 306.0.2+up0.7.0
+:version-endpoint-copier-operator-chart: 306.0.1+up0.3.0
+:version-kubevirt-chart: 306.0.2+up0.7.0
+:version-kubevirt-dashboard-extension-chart: 306.0.4+up1.3.3
+:version-longhorn-chart: 1.11.1
+:version-longhorn-crd-chart:
```
diconico07: This one is no longer needed I believe

diconico07 (dismissed):
A few changes related to recent changes to the docs to reflect in the template, otherwise LGTM

asciidoc/edge-book/versions.adoc
hardys: Thanks, removed

asciidoc/edge-book/versions.adoc
hardys: Thanks, added logic to the script to derive this from the metal3 chart values

scripts/README.md
```diff
@@ -0,0 +1,13 @@
...
+These scripts are used to simplify the maintenance of the documentation in this repo
+
+
+
+## Update release versions
+
+The `versions_update.sh` script updates the `versions.adoc` using data from the Factory `release_manifest.yaml`
+
+```bash
+./versions_update.sh
```
diconico07:
```suggestion
The `versions_update.py` script updates the `versions.adoc` using data from the Factory `release_manifest.yaml`

```bash
./versions_update.py
```

asciidoc/edge-book/versions.adoc.j2
```diff
@@ -0,0 +1,144 @@
...
+
+// == Edge Image Builder ==
+:version-eib: {{ version_eib }}
+:version-eib-api-latest: 1.3
+
+// KubeVirt
+//  This is used in download URLs and filenames from upstream, so it must have
+//  the leading "v". If needed, a separate version-kubevirt should be created
+//  with simply the version number itself.
+:version-kubevirt-release: v{{ release_manifest_kubevirt_version .split("+up") | last }}
```
diconico07:
```suggestion
:version-kubevirt-release: v{{ release_manifest_kubevirt_version.split("+up") | last }}
```

diconico07 (commented):
LGTM, just those two suggestions caught by AI that I missed initially.

alknopfler (approved):
LGTM good job

PR #1018, opened 2026-01-20
releasenotes: Add a note for ptp4l configuration

mkrutov-suse:


asciidoc/edge-book/releasenotes.adoc
```diff
@@ -64,6 +64,20 @@ Summary: SUSE Edge 3.5.0 is the first release in the SUSE Edge 3.5 release strea
 == Known Issues
 
 * The SUSE Multi-Linux Manager version included with SUSE Edge 3.5 (5.0.6) doesn't support SUSE Linux Micro 6.2 yet. It will be updated and supported in future SUSE Edge 3.5 releases.
+* Running https://documentation.suse.com/suse-edge/3.4/html/edge/atip-features.html#ptp-telco-config[PTP feature] in default configuration requires additional steps:
+Following needs to be added to overrides in the ptp4l service for it to work correctly in Edge 3.5 release:
```
hardys: I think the URL here should use the `{version-edge}` attribute instead of linking to the 3.4 docs?

And small wording nit - `The following needs to be added to overrides in the ptp4l service for it to work correctly in the Edge 3.5 release`

hardys (approved):

PR #1032, opened 2026-02-02
Addresses 1031 - Update configure-network.sh

### LINKED ISSUE #1031: Update configure-network.sh in quickstart/metal3.adoc to reflect telco-cloud-examples
The script configure-network.sh has been updated in the telco-cloud-examples to work in scenarios where there is no network_data.json file. This needs to be correctly reflected in the documentation as well. 

For more details see
```bash
./edge-clusters/airgap/eib/network/configure-network.sh
./edge-clusters/dhcp-less/eib/network/configure-network.sh
./edge-clusters/dhcp/eib/network/configure-network.sh
./edge-clusters/aarch64/eib/network/configure-network.sh
```

ehsjoar:
This is to address situations where we don't get any network data but still want to set the hostaname (e.g. in the case of Sylva). This MR fixes #1031

### DIFF
--- a/asciidoc/product/atip-automated-provision.adoc
+++ b/asciidoc/product/atip-automated-provision.adoc
@@ -289,6 +289,16 @@ fi
 
 mount -o ro $CONFIG_DRIVE /mnt
 
+META_DATA_FILE="/mnt/openstack/latest/meta_data.json"
+if [ ! -f "${META_DATA_FILE}" ]; then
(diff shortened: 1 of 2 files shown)


hardys (approved):

hardys:
Merging this so we can make progress - if there are further changes desired lets handle that via a follow-up PR

PR #1107, opened 2026-04-10
Update ironic ConfigMap name

fdegir:
Ironic ConfigMap, ironic-bmo, has been renamed to ironic so this commit updates the troubleshooting documentation with the updated ConfigMap name.


fdegir:
Needs to be cherry-picked to 3.5 as well.

diconico07 (approved):
Nice catch, we definitely missed to update this back then 👍

PR #1126, opened 2026-05-06
Improve the PHC to system clock synchronization section

mchiappero:
The existing [42.13.2.3 Configuration of phc2sys](https://documentation.suse.com/suse-edge/3.5/html/edge/atip-features.html#id-configuration-of-phc2sys) isn't particularly refined, expand and improve wording.

PR #1129, opened 2026-05-07
telco-features: include a section for configuring a PTP Boundary Clock

mchiappero:
The initial documentation for Precision Time Protocol only includes instruction on how to configure an Ordinary Clock (for the Time Consumer use case).

Most Telco deployments require instead the node to behave as a Boundary Clock, so extend the documentation to include a section about ptp4l as BC.

### DIFF
--- a/asciidoc/product/atip-features.adoc
+++ b/asciidoc/product/atip-features.adoc
@@ -1,6 +1,6 @@
 [#atip-features]
 == Telco features configuration
-:revdate: 2026-05-07
+:revdate: 2026-05-08
 :page-revdate: {revdate}
 :experimental:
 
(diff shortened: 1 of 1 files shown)


asciidoc/product/atip-features.adoc
```diff
@@ -1888,6 +1884,58 @@ Check the synchronization status by observing the logs with:
 # journalctl -e -u ptp4l
 ----
 
+===== PTP configuration of a Boundary Clock
+
+The previous sections covered the configuration of an Ordinary Clock, but Telco deployments can require the node to act as a Boundary Clock. In such a scenario the reference time must be propagated from a Grand Master to one or more Time Receivers downstream, which requires a NIC with two or more ports. In this case one port will take the Time Receiver role, that is driving the updates of the PTP Hardware Clock (PHC) on the NIC, while the remaining ones will work as Time Transmitters. A single instance of ptp4l is sufficient for a single multi-port NIC, but each kernel interface involved in the time distribution must be provided, as part of its configuration. For example, `/etc/sysconfig/ptp4l` can be updated to include four ports, just replace the variables with actual interface names or adjust it as needed:
```
ranjinimn:
```suggestion
The previous sections covered the configuration of an Ordinary Clock, but Telco deployments can require the node to act as a Boundary Clock. In such a scenario the reference time must be propagated from a Grand Master to one or more Time Receivers downstream, which requires an NIC with two or more ports. In this case one port will take the Time Receiver role, that is driving the updates of the PTP Hardware Clock (PHC) on the NIC, while the remaining ones will work as Time Transmitters. A single instance of `ptp4l` is sufficient for a single multi-port NIC, but each kernel interface involved in the time distribution must be provided, as part of its configuration. For example, update the  `/etc/sysconfig/ptp4l` file to include four ports, just replace the variables with actual interface names or adjust it as needed:
```

asciidoc/product/atip-features.adoc
```diff
@@ -1888,6 +1884,58 @@ Check the synchronization status by observing the logs with:
...
+
+[,shell]
+----
+OPTIONS="-f /etc/ptp4l-G.8275.1.conf --message_tag ptp-8275.1 -i $IFNAME1 -i $IFNAME2 -i $IFNAME3 -i $IFNAME4"
+----
+
+Note, there is no need to set `clock_type` to `BC`, as it will be automatically implied if more than one port is provided.
```
ranjinimn:
```suggestion
[NOTE]
====
There is no need to set `clock_type` to `BC`, as it will be automatically implied if more than one port is provided.
====
```

asciidoc/product/atip-features.adoc
```diff
@@ -1888,6 +1884,58 @@ Check the synchronization status by observing the logs with:
...
+serverOnly                      1
+
+[$IFNAME3]
+serverOnly                      1
+
+[$IFNAME4]
+serverOnly                      1
+----
+
+Make sure however to determine the correct connection to the GM before starting the daemon.
```
ranjinimn:
```suggestion
However, make sure to determine the correct connection to the GM before starting the daemon.
```

asciidoc/product/atip-features.adoc
```diff
@@ -1888,6 +1884,58 @@ Check the synchronization status by observing the logs with:
...
+
 ===== Configuration of phc2sys
```
ranjinimn: Should this be `phc2sys` or just phc2sys? 

```suggestion
===== Configuration of `phc2sys`
```

ranjinimn (commented):
@mchiappero I had a few suggestions. I would be happy to review again :) Thanks.

alknopfler (dismissed):
From a technical point of view the content LGTM Marco! Thanks

hardys:
lgtm but needs a rebase due to #1126 merging

mchiappero:
> @mchiappero I had a few suggestions. I would be happy to review again :) Thanks.

Thanks a million, all applied!

mchiappero:
> lgtm but needs a rebase due to #1126 merging

Done!

hardys (approved):

PR #1131, opened 2026-05-12
metal3: document airgap IPA image configuration with hauler

diconico07:
When using EIB/hauler to install metal3 in airgap environments, the ironic-python-agent (IPA) image must be:
- Added to the embedded artifact registry images list
- Configured to use hauler via metal3-ironic.ipa.useHauler: true

Changes:
- Add version-ipa attribute (3.0.8) to versions.adoc for templating
- Add Metal3 Installation section to air-gapped-eib-deployments.adoc with configuration snippets and useHauler requirement
- Update atip-management-cluster.adoc to include IPA image in embeddedArtifactRegistry and useHauler in metal3.yaml values

### DIFF
--- a/asciidoc/edge-book/versions.adoc
+++ b/asciidoc/edge-book/versions.adoc
@@ -36,6 +36,7 @@
 :version-nvidia-device-plugin: 0.14.5
 :version-kiwi-builder: 10.2.29.1
 :version-private-registry: 1.1.1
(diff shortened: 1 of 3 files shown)


hardys (approved):
It's unfortunate we need a new flag for this, but I guess there's no easy way for us to detect in the chart when we're deploying via EIB 🤔

PR #1097, opened 2026-04-02
Split Edge and Telco documentation

fsundermeyer:
Splits Edge and SUSE Telco Cloud docs to enable generation of independent docs builds with overlapping but in some cases different content

PR #1135, opened 2026-05-18
Rework Telco structure

hardys:
After merging #1097 we now have separated docs for Edge and Telco, this aims to iterate on that split and pull the previously nested telco content into the top-level TOC


hardys:
Thanks @jiwonhu for the suggested structure

alknopfler (approved):
I've checked it out with the preview you shared @hardys 
Looks much much better now. I love the idea to move down the how-to guides because some customers reported the feedback just based on those section guides, so now our product docs are better organized

PR #1143, opened 2026-05-27
Clarify net.ifnames default and directed network provisioning model

jiwonhu:
Raising a new PR as suggested by @hardys in #1136 

- Fixes #1132 by documenting `net.ifnames=1` should be enforced for SLM<6.2. Previous versions of this doc has a wrong value and this has to be backported
- The previous "Example deployment flows" section framed the deployment model as examples, which could be understood as other kinds of deployment models are supported. It's re-written as a "Deployment Model" section that clarifies Directed Network Provisioning is the supported method for downstream clusters, and explains "why", which I believe is important but was lacking in the doc.

PR #1150, opened 2026-05-27
Cherry-pick-3.4-Clarify net.ifnames default and directed network provisioning model

ranjinimn:
Closes https://github.com/suse-edge/suse-edge.github.io/pull/1143


hardys:
@ranjinimn I think this is a duplicate of #1167 ?

PR #1151, opened 2026-05-27
Cherry-pick-3.3-Clarify net.ifnames default and directed network provisioning mode

ranjinimn:
Closes https://github.com/suse-edge/suse-edge.github.io/pull/1143


ranjinimn:
@hardys The build is failing because it is looking for a section "CPU Pinning on Kubernetes". Should I include it in this and the earlier versions?

hardys:
> @hardys The build is failing because it is looking for a section "CPU Pinning on Kubernetes". Should I include it in this and the earlier versions?

If https://github.com/suse-edge/suse-edge.github.io/pull/976 can be cleanly backported we could add it, but IMO it's not worth spending a lot of time backporting this far if significant effort is required to resolve conflicts

PR #1158, opened 2026-06-01
Edge1984 2086 2113

antaloala:
Latest updates required for SUSE 3.6.0 docs, three different commits for each:

- 1st commit: Updates "ACC100/ACC200" section in SUSE Telco Cloud doc, adding the new VRB2 support (for GNR-D) and the now needed "-f" setting when invoking the "pf_bb_config" command (required from the latest "pf-bb-config" rpm supporting the VRB2 device type needed for GNR-D support). Old "ACC200" term now replaced by the new "VRB1" one
- 2nd commit: Updates the name of the kiwi profiles referred to generate the downstream EIB base "raw" images, using now "Base" and "Base-RT·(instead of "Base-SelfInstall" and "Base-RT-SelfInstall"  which are used to generate "iso" and not "raw" images)
- 3rd commit: Adds the sctp and vfio-pci config files to downstream EIB folder (in the os-folder path) & then removes the not needed anymore "modprobe vfio-pci" command from .spec.preRKE2Commands stanza from RKE2ControlPlane and RKEConfigTemplate manifests


alknopfler (dismissed):
LGTM

alknopfler (approved):

PR #1161, opened 2026-06-02
Update MLM version

hardys:
This was missed when bumping versions for 3.6


renner:
Thanks, as mentioned elsewhere: MLM 5.1.3 might be the latest 5.1 release at the time Edge 3.6 is released, but we are currently doing an update more or less every two months. I'm not really sure how you would ideally reflect this, but I guess the idea is not that Edge customers would stick to 5.1.3 during the lifecycle of Edge 3.6. So maybe it could be an option at least in the future to either just point to MLM 5.1, expecting people to always upgrade to the latest, or somehow make clear that it is 5.1.3 or newer?

hardys:
> Thanks, as mentioned elsewhere: MLM 5.1.3 might be the latest 5.1 release at the time Edge 3.6 is released, but we are currently doing an update more or less every two months. I'm not really sure how you would ideally reflect this, but I guess the idea is not that Edge customers would stick to 5.1.3 during the lifecycle of Edge 3.6. So maybe it could be an option at least in the future to either just point to MLM 5.1, expecting people to always upgrade to the latest, or somehow make clear that it is 5.1.3 or newer?

Thanks for the feedback @renner!  We do semi-regular z-stream releases which can include newer patch versions of MLM and other components, but right now MLM isn't fully integrated into our test automation, so if we're in agreement it may make sense to just align with a specific y-stream e.g 5.1 - @joesusecom any thoughts?

hardys:
Now updated to only specify the y-stream version - @joesusecom if you're OK with this we can merge/backport

joesusecom:
Fine with this. We just need to remember to actually do a manual 5.1.x validation soon.

dprodanov4 (approved):

PR #1163, opened 2026-06-08
scripts: use ORAS for manifest and config-blob fetches

hardys:
registry.suse.com requires a bearer token even for anonymous pulls,
which the bare requests.get() calls did not negotiate. Route the
manifest and config-blob fetches through client.get_manifest() and
client.get_blob() so ORAS handles the token exchange.

Requires oras>=0.2.41 (first release with manifest-list-aware
defaults and opt-in schema validation); enforced at import time
with a clear error.

Tested with:
  `./versions_update.py --image registry.suse.com/edge/3.6/release-manifest --tag 3.6.0`

### DIFF
--- a/scripts/versions_update.py
+++ b/scripts/versions_update.py
@@ -4,6 +4,7 @@ Generate versions.adoc from release manifest in container image.
 Uses ORAS library for OCI registry interaction.
 """
 
+import oras
 import oras.client
 import requests
 import tempfile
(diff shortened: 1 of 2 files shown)


scripts/versions_update.py
```diff
@@ -101,6 +101,38 @@ def find_file_in_layer(layer_path, target_file):
     return None
 
 
+def _request_with_bearer_auth(url, headers=None):
```
diconico07: Might be better to just use the ORAS library to download the manifest then

hardys: Unfortunately ORAS only covers part of the flow — client.pull() [doesn't resolve image indexes ](https://github.com/oras-project/oras-py/issues/108). The script therefore has to fetch the index, pick a platform, re-fetch the platform manifest, and pull the config blob itself — and those raw HTTP calls are what need the bearer-token helper.

diconico07: `client.get_manifest("registry.suse.com/edge/3.6/release-manifest:3.6.0")` returns the index actually

hardys: Yes but it won't pick a platform and re-fetch - so the script has to work around that for multi-arch images

Perhaps I can find a way to make better use of the get_manifest index data, but I think we'll still have to do some raw requests unless changes are made inside the oras library?

hardys: Ah I was missing https://github.com/oras-project/oras-py/pull/237 in my dev environment, after updating it is possible to simplify things and use more of the oras functions

diconico07 (approved):

PR #1164, opened 2026-06-09
Adds network parameters section to document tuning for optimal perfor…

alknopfler:
This pull request updates the `asciidoc/product/atip-features.adoc` documentation to introduce and detail network tuning parameters for optimal performance in telco workloads. It adds a new section describing the network parameters configured by the `performance-settings.sh` script and updates references throughout the document to reflect these enhancements.

**Documentation updates for network tuning:**

* Added a new section, `Network Parameters`, describing the purpose and configuration of various kernel network parameters (e.g., buffer sizes, UDP settings, real-time scheduling) set by the `performance-settings.sh` script for telco workloads.
* Updated the section overview to include `Network Parameters` as a covered topic.
* Clarified that the tuning script now configures both CPU and network parameters for optimal performance.…mance in telco workloads

### DIFF
--- a/asciidoc/product/atip-features.adoc
+++ b/asciidoc/product/atip-features.adoc
@@ -23,6 +23,7 @@ The following topics are covered in this section:
 * <<kernel-args,Kernel arguments for low latency and high performance>>: Kernel arguments to be used for maximum performance and low latency running telco workloads.
 * <<cpu-pinning-host,CPU Pinning on Host>>: Isolating the CPUs via kernel arguments and TuneD profile.
 * <<cpu-pinning-k8s,CPU Pinning on Kubernetes>>: Isolating the CPUs on Kubernetes via Kubelet configuration.
+* <<network-parameters,Network Parameters>>: Network tuning parameters for optimal performance and low latency.
 * <<cni-configuration,CNI configuration>>: CNI configuration to be used by the Kubernetes cluster.
 * <<sriov,SR-IOV configuration>>: SR-IOV configuration to be used by the Kubernetes workloads.
 * <<dpdk,DPDK configuration>>: DPDK configuration to be used by the system.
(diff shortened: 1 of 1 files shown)


alknopfler:
@antaloala keep in mind we're not including the content of the performance because we've got a link directly to the telco examples repo, so the only modifications with the PR is the param explanation in telco features as we talked.

asciidoc/product/atip-features.adoc
```diff
@@ -350,6 +352,55 @@ spec:
...
+
+* `net.core.netdev_max_backlog` (416384 packets): Maximum number of packets queued on the input side when the interface receives packets faster than the kernel can process them. Increasing this value helps prevent packet drops during traffic bursts.
+
+* `net.core.optmem_max` (25165824 bytes): Maximum ancillary buffer size allowed per socket. This affects the size of option memory buffers.
+
+* `net.core.netdev_budget` (1024 packets): Maximum number of packets that can be processed in one softirq run. This helps control CPU utilization during packet processing.
+
+=== UDP-Specific Parameters
+
+* `net.ipv4.udp_mem` ("11416320 15221760 22832640"): Controls the amount of memory (in pages) used by UDP. The three values represent:
```
antaloala: I would remove the " ... " surrounding the three values as it look to invite to set this parameter using those, but that makes this `sysctl -w net.ipv4.udp_me` command to fail (I experienced it in the past and it took me some time to realize on it :-(
As in the other parameters you are giving the "unit" used for each value, why not to do the same for this?

E.g. `net.ipv4.udp_mem` (11416320 bytes, 15221760 bytes, 22832640 bytes):

alknopfler: changed... the units are pages so I added but it's not the same than others as you can imagine ;)

antaloala: Ops, you are right (they are pages but not bytes ;-)

antaloala (approved):
lgtm

PR #1168, opened 2026-06-12
Updated release notes for kernel bug

ehsjoar:
Updated release doc to include a known issue about hung system when using SR-IOV virtual functions in pods


alknopfler (approved):
LGTM

PR #1196, opened 2026-07-17
Fs/3.1 add revision number

fsundermeyer:
Asciidoctor only creates a revision history when both revdate and revnumber are set. A revision history is required to set a correct modified date for the file, which in terms is a must for SEO, SearchUnify and AskGeeko.


fsundermeyer:
Wrong merge target, ignore

PR #1205, opened 2026-07-28
Add release documentation update generators

dprodanov4:
Adds manifest-driven documentation generators for z-stream and minor SUSE Edge releases.

The scripts accept release manifest container images, local manifest directories, raw manifest URLs, and Factory Git refs. They update deterministic release data while leaving release-note prose, lifecycle information, CVEs, known issues, and other non-manifest documentation for manual review.

Validation:
- Python compilation and tabnanny checks
- z-stream replay tests for 3.4.4, 3.5.2, and 3.6.1
- minor-release replay against the older 3.4 documentation layout
- idempotence checks after generated updates

### DIFF
--- a/scripts/generate_zstream_release_updates.py
+++ b/scripts/generate_zstream_release_updates.py
@@ -1,0 +1,907 @@
+#!/usr/bin/env python3
+"""Generate SUSE Edge documentation updates for a z-stream release.
+
+The release manifest container is the source of truth. The script consumes the
+three files shipped in that container:
+
+* release_manifest.yaml
+* release_images.yaml
+* tooling_manifest.yaml
+
+It updates version attributes, known air-gap image examples, and an
+already-created release-note component table. Narrative release notes,
+lifecycle dates, CVEs, and image-list additions/removals remain manual review
+items.
+"""
+
+from __future__ import annotations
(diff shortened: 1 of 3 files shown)


scripts/generate_zstream_release_updates.py
```diff
@@ -0,0 +1,907 @@
...
+    for candidate in candidates:
+        if candidate and shutil.which(candidate):
+            return candidate
+    raise RuntimeError(f"could not find container engine: {requested or 'podman or docker'}")
+
+
+def ensure_container_image(image: str, engine: str, skip_pull: bool) -> None:
+    if not skip_pull:
+        print(f"Pulling release manifest image: {image}", file=sys.stderr)
+        subprocess.check_call([engine, "pull", image])
```
hardys, line 174: IMO it would be better to use the approach from https://github.com/suse-edge/suse-edge.github.io/blob/main/scripts/versions_update.py - this uses the ORAS client to pull the images without requiring any local container runtime which makes the script a lot more portable.

Also I wonder if we can perhaps abstract the logic from versions_update.py into another file then have a common function to e.g get the release manifest so we don't duplicate that logic in each script?

PR #1222, opened 2026-09-04
Migrate Metal3 examples to the supported provider-id format

hardys:
Align the Metal3 examples with telco-cloud-examples#106, which migrates away from the deprecated provider-id format:

```
  old (deprecated): metal3://<BMH-UID>
  new (supported):  metal3://<namespace>/<bmh-name>/<m3m-name>
``` 

The legacy format was deprecated in CAPM3 v1.13 (3.7 version) and is will be removed for 3.8 in v1.14, see:

  https://github.com/metal3-io/metal3-docs/blob/main/docs/user-guide/src/capm3/providerid-workflow.md

The `metal3.io/uuid` node label is intentionally left unchanged - it still carries the BareMetalHost UUID and is required by CAPM3 to match the Nodes to Machines.

### DIFF
--- a/docs/en/guides/clusterclass.adoc
+++ b/docs/en/guides/clusterclass.adoc
@@ -213,7 +213,8 @@ spec:
                 Type=oneshot
                 User=root
                 ExecStartPre=/bin/sh -c "mount -L config-2 /mnt"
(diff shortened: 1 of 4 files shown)


alknopfler (approved):
LGTM thanks for the improvements

PR #1224, opened 2026-09-04
3.7 docs

dprodanov4:


docs/en/guides/clusterclass.adoc
```diff
@@ -666,14 +666,14 @@ spec:
...
+    version: v1.36.3+rke2r1
     controlPlane:
       replicas: 1
     variables:                       # Values to be replaced in the clusterclass template variables to create this specific cluster
     - name: endpointCopierOperatorHelmChartVersion
-      value: 306.0.1+up0.3.0
+      value: 307.0.1+up0.3.0
     - name: metallbHelmChartVersion
-      value: 306.0.2+up0.15.3
+      value: 307.0.3+up0.16.1
```
hardys, line 676: It would be better if we updated these hard-coded references to attributes instead?

docs/en/product/atip-automated-provision.adoc
```diff
@@ -582,7 +582,7 @@ With SUSE Private Registry:
 ----
 $ podman run --rm \
   -v ./:/opt:z \
-  registry.suse.com/edge/{version-edge-registry}/release-manifest:{version-edge} \
+  registry.suse.com/edge/{version-edge-registry}/release-manifest:{version-edge}.0 \
```
hardys, line 585: We should avoid using a hard-coded suffix here, can we add a version-release-manifest attribute instead so we don't forget to update this for z-stream releases?

docs/en/tips/metal3.adoc
```diff
@@ -173,7 +173,7 @@ Create the following script at `/opt/EIB/network/configure-network.sh`:
 #!/bin/bash
 set -eux
 
-# Source: https://documentation.suse.com/suse-edge/3.6/html/edge/quickstart-metal3.html#metal3-add-network-eib
+# Source: https://documentation.suse.com/suse-edge/3.7/html/edge/quickstart-metal3.html#metal3-add-network-eib
```
hardys, line 176: We could use the `version-edge` attribute here instead of the hard-coded value

scripts/generate_zstream_release_updates.py
```diff
@@ -0,0 +1,909 @@
...
+    return parsed.isoformat()
+
+
+def load_yaml(path: Path) -> Any:
+    with path.open("r", encoding="utf-8") as handle:
+        return yaml.safe_load(handle)
+
+
+def find_container_engine(requested: str | None) -> str:
+    candidates = [requested] if requested else ["podman", "docker"]
```
hardys, line 164: As mentioned in https://github.com/suse-edge/suse-edge.github.io/pull/1205 it would be really nice to avoid requiring a container engine just to pull the image - see https://github.com/suse-edge/suse-edge.github.io/blob/main/scripts/versions_update.py 

Perhaps you could get the agent to refactor so we use common code/approach?

Not necessarily blocking on that but IMO it would be nice to avoid fragmentation and additional dependencies on the host running the script.

hardys (commented):
Overall looks good - I have a few comments, let me know if you'd like to address those now or as a follow-up

dprodanov4:
hey Steve, thanks for the review. I fixed all the issues and it is ready for review again

hardys (approved):
The script implementation looks better but IMHO still not super clean, but lets iterate on that in a follow-up as I don't want to block progress.

alknopfler (approved):

dprodanov4:
thanks Steve, you are right we can optimise the script in future PRs

PR #1226, opened 2026-09-08
moved README file to the root folder

ranjinimn:
Moved README.adoc file ti the root folder as per the discussion over Slack.


fsundermeyer (changes requested):
Now we have two conflicting README files in the root folder. I suggest to move this README to docs/ instead. Also, the link to the local build instructions in README.md in the root folder will need to be adjusted.

fsundermeyer (approved):
LGTM, thank you!

PR #1232, opened 2026-09-17
Various fixes required for clean translations:

fsundermeyer:
* added missing IDs
* changed auto-IDs to fixed IDs
* disabled auto IDs
* fixed an escaping issue ${IP} -> $\{IP}

**A note on auto IDs:**

Asciidoc generates IDs when converting to XML from headlines or titles if they do not exist:

== My title

gets the ID `id-my-title` which you can use in links (xrefs or <<>>). However, since these auto-IDs include the title string, they change when a document gets translated. In a German version of the document the ID would become `id-mein-titel` and links to `id-my-title` will no longer work. For some Asian languages this will result in IDs containing 8-bit characters, which are not valid for XML IDs.

Therefore I disabled auto IDs with this PR. As a result, every element you want to link to needs to have a proper unique ID. For all elements that do not have an ID set in asciidoc, no ID will be generated in XML as well. When converting to HTML, this will result in randomly generated IDs (which usually is accepatble). However note that each toplevel headline MUST have an ID in asciidoc.

**A note on "direct" links**

asciidoctor allows to directly link to a title when using the <<>> syntax:

```
==My title

See <<My title>> for more information.
```

Apparently this results in a broken link as well when the document is translated. This feature MUST no longer be used.


ranjinimn (approved):

hardys (approved):
lgtm, thanks!

PR #1216, opened 2026-08-06
Moving adoc files into new directory structure...

fsundermeyer:
...required for translation with Phrase.
Rolling out the change, that was already done for release-3.6 a while ago, to main.

PR #1233, opened 2026-09-17
Fix workflow paths

hardys:
This was missed in #1216 - we need to adjust this or the publish to https://suse-edge.github.io/ is always disabled with the new directory structure

Also remove stale paths which are no longer needed


ranjinimn (approved):
Good catch. looks good :)

PR #1239, opened 2026-09-21
releasenotes: Updates for 3.7

hardys:
Update the release dates and digest for the sbom example

PR #1244, opened 2026-09-22
[3.7] backports (#1237 #1239 #1240)

hardys:
Backport #1237 #1239 #1240 which recently landed on main and are needed for 3.7


antaloala (approved):
lgtm

